PR #3, opened 2020-12-02
Fix #2: use high precision in the fragment shader so that linear gradients render correctly on NVidia hardware.

### LINKED ISSUE #2: Linear gradients render incorrectly on NVIDIA/Linux
I noticed that linear gradients aren't rendered correctly on NVIDIA on Linux (but Intel graphics on Linux works fine). I believe the problem is because the linear case uses extents for the signed distance round-rect function in the fragment shader which are on the edge of `mediump`'s precision, causing an unexpectedly lumpy interpolation.

If I use `highp` precision in the fragment shader, then NVIDIA renders as expected (and the same as Intel).

iamralpht:
The linear gradient code uses extents with small deltas around 1e5, and this causes lumpy interpolation in the sdRoundRect function used to calculate gradients. Box gradients and radial gradients don't use smaller extents and don't have the same issue.

### DIFF
--- a/src/renderer/opengl/main-fs.glsl
+++ b/src/renderer/opengl/main-fs.glsl
@@ -1,5 +1,5 @@
 
-precision mediump float;
+precision highp float;
 
 #define UNIFORMARRAY_SIZE 12
 


adamnemecek:
Thanks!

PR #6, opened 2020-12-15
Switch from Harfbuzz to Rustybuzz

tronical:
The latest release of Rustybuzz is now a pure Rust port, without any
external C/C++ dependency. Using it here paves the way for a wasm build,
for example.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,15 +131,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cmake"
-version = "0.1.44"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


tronical:
This compiles and on my macOS install I can't see any difference in text rendering for the text example. But is this the right approach? Is this desired at all? :)

adamnemecek:
Ok how about this. Keep using this and I’ll merge it a couple of days assuming no problems emerge.

tronical:
Sure, no hurry from my side. Do you have any good smoke tests that would help locate potential issues?

adamnemecek:
Ok, now that I know that you intend to use it for sixtyfps (how certain are you about this?), I'm willing to merge it in. I guess I was trying to avoid a scenario where I'd merge it in, and then discover some strange text issues. If you are going to keep using it, then I guess you will discover said problems.

Issue #1, opened 2020-11-09
Is Web Supported/Planned?


zicklag:
Hey there, this looks pretty nice. :) I was curious if it is possible to run in the browser. Maybe [`glow`](https://docs.rs/glow) could be useful for that if it isn't supported already.

zicklag:
Not sure I'll get to this, but would you be open to a PR to port the OpenGL renderer to `glow`? I'm really liking the look of this library so far and I am considering using it for an application I am making that needs text page rendering, but needs to be able to run on web.

adamnemecek:
Hey, sorry about the delayed response. I'm planning to add wgpu support but probably not anytime soon. I'm working on a metal backend https://github.com/adamnemecek/gpucanvas (the project used to be called gpucanvas).

I'm very open to a glow backend but for starters add it as another backend for now as opposed to porting the current backend.

zicklag:
OK, cool. I'm still looking around so I may not get to this, but I wanted to make sure it was an option if this is what I wanted to try. :)

adamnemecek:
Are you aware of wgpu?

zicklag:
Yes, but it doesn't yet support WebGL. I'm actually one of the ones who was working on getting GL support in WGPU. :wink:

Its getting there, but not quite ready yet.

adamnemecek:
Ah, good luck with that.

tronical:
I'm kind of in the same boat -- I'd be interested in a GLES 2.0 renderer support based on glow, for WebGL 1 (since that's the only thing that works on iPhones out of the box in mobile Safari).

Would patches that go into this direction find likely acceptance? :)

What are your thoughts with regards to GLES 2.0 support?

P.S.: Great project, btw!

adamnemecek:
Do you want to take a stab at it? I’ll be honest, right now the plan is to be adding support for the new APIs (metal, wgpu) however feel free to add support for GLES.

tronical:
I might be interested in that yes - I see an opportunity to replace our own renderer with femtovg :).

What are your thoughts with regards to adding features to the API that would end up being difficult to support on GLES?

adamnemecek:
Which features are you referring to? What’s your intended use case? What are you using now?

tronical:
I don't know what kind of features in particular would be difficult to support on GLES 2.0.

[...]

PR #7, opened 2020-12-17
Port to glow as GL abstraction

tronical:
This request includes two changes, a port to glow (please review carefully!) and a small change on top to enable a wasm build and make it possible to instantiate the GL renderer accordingly.

If desired I can also split these commits into separate pull requests.


tronical:
Forgot to link the corresponding issue, this merge request relates to issue #1

tronical:
I noticed that I needed to do more fine-tuning on the WASM API/interface, so it makes more sense to split requests. So this merge requests *only* contains the port to glow and once that's merged then I'll prepare another merge request with a series of changes for the WASM port.

PR #9, opened 2020-12-21
Bump glow version

tronical:
This allows removing the Cargo patch


peter-todorov:
Thanks Tronical! Glow updated that fast.

Issue #10, opened 2020-12-21
Text doesn't render correctly when the inner window size is too small


geom3trik:
If the inner window size is set too small text either renders incorrectly or completely dissapears. For example, on the Demo example, setting the height to 500 causes some characters to dissapear, like lowercase 'e', and setting the height to 400 causes all text to dissapear. The same is true for width but not with the same values.

adamnemecek:
@geom3trik thanks for reporting this. want to take a stab at fixing it?

geom3trik:
I can give it a go but I'm not really sure how to even go about debugging this. Have you got any ideas what might be causing it?

adamnemecek:
What os are you using? There are some debuggers for opengl like nvidia nsight or renderdoc.

geom3trik:
I am on windows. I'll give it a go but this is a bit out of my area.

adamnemecek:
I'll take a peek in a bit, there is a bunch of stuff I need to fix and I'll batch it together.

PR #11, opened 2020-12-22
Fix for #10 - wrong text rendering on small window sizes

peter-todorov:
canvas.width() and canvas.height() were returning wrong values when rendering to the texture (ref #10)


adamnemecek:
You should have permissions to merge in stuff now. Sorry about that.

PR #14, opened 2020-12-25
opengl: support rendering to an externally-created framebuffer object.

iamralpht:


iamralpht:
Usage would be something like
```
    let mut renderer = OpenGl::new(load_fn)?;
    renderer.set_screen_target(target_fbo);
    let canvas = Canvas::new(renderer)?;
```

peter-todorov:
Awesome, thanks Ralph!

PR #15, opened 2020-12-27
feature: add multi-stop gradients

iamralpht:
This change implements multi-stop gradients, and works for basic cases. It adds a gradient store, which is a cache of gradient textures maintained by the renderer. It might be that we want the Canvas to maintain the gradient cache.

### DIFF
--- a/src/gradient_store.rs
+++ b/src/gradient_store.rs
@@ -1,0 +1,150 @@
+use std::collections::BTreeMap;
+
+use imgref;
+use rgb;
+
+use crate::{Color, ErrorKind, ImageFlags, ImageId, ImageInfo, ImageSource, Renderer, image::ImageStore, paint::MultiStopGradient};
+
+/// GradientStore holds image ids for multi-stop gradients. The actual image/textures
+/// are contained by the Canvas's ImageStore.
+//
+// If many gradients are used in a frame, we could combine them into a single texture
+// and update the texture immediately prior to giving the renderer the command list.
+pub(crate) struct GradientStore {
+    this_frame: BTreeMap<MultiStopGradient, ImageId>,
+    prev_frame: BTreeMap<MultiStopGradient, ImageId>
+}
+impl GradientStore {
+    /// Create a new empty gradient store
+    pub fn new() -> GradientStore {
+        GradientStore {
+            this_frame: BTreeMap::new(),
+            prev_frame: BTreeMap::new()
+        }
+    }
+
+    /// Lookup or add a multi-stop gradient in this gradient store.
+    pub fn lookup_or_add<R: Renderer>(&mut self, colors: MultiStopGradient, images: &mut ImageStore<R::Image>, renderer: &mut R) -> Result<ImageId, ErrorKind> {
+        if let Some(gradient_image_id) = self.prev_frame.remove(&colors) {
+            // See if we already have this texture from the previous frame. If we find
+            // it then we migrate it to the current frame so we don't release it and
+            // return the texture id to the caller.
+            self.this_frame.insert(colors, gradient_image_id);
+            Ok(gradient_image_id)
+        }
+        else if let Some(gradient_image_id) = self.this_frame.get(&colors) {
+            // See if we already used this gradient in this frame, and return the texture
+            // id if we do.
+            Ok(*gradient_image_id)
+        } else {
+            // We need to allocate a texture and synthesize the gradient image.
+            let info = ImageInfo::new(
+                ImageFlags::REPEAT_Y,
+                256,
+                1,
+                crate::PixelFormat::Rgba8,  
+            );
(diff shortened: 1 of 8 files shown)


iamralpht:
This change works but is incomplete (no box variation, no tests, I bet I'm not applying global alpha or premultiplying etc).

The main questions here are:
1. Is it OK for the gradient cache to live in the Renderer? We could move it to the Canvas and potentially pack multiple gradients into a single texture, tho we'll want to be careful to not update a texture after we've submitted a command that uses it.
2. What should the public API look like?
3. What should happen to `Params`? We can make it contain an enum for the different color types, or if we kept the gradient store in the Canvas (and came up with an id scheme for gradients) we could either store the gradient id in `Params` or pass through the gradient info as we pass through the image id...

tronical:
Cool!

Do I understand this patch correctly that the design is similar to the nanovg upstream PR https://github.com/memononen/nanovg/pull/430 ?

iamralpht:
@tronical yep, multistop gradients cause the renderer to make a 256x1 texture containing the gradient, then the fragment shader samples from that texture instead of mixing between the inner color and outer color. I'm not sure if that PR against nanovg supports arbitrarily positioned stops (the first change says the stops are equidistant) but this change does support arbitrary stop positions. It's not like the Rodrigo change in the comments of the PR because we don't evaluate the gradient color in the shader (because I suspect it would be slow, but I haven't measured, and I suspect we'd hit uniform count limits on GLES2, but I haven't checked).

I think the shader change I have should be fine, the big question is what the best way to move gradients from the canvas to the renderer. The way I've done it here is a small change, but I'm suspecting that it'd be better to make `GradientStore` belong to `Canvas` and be friends with `ImageStore`, and make the gradient textures work more like synthesized images. This is a much bigger change to `Canvas` but it avoids all the ugliness with `Params`.

The other choice I made that's a bit dubious is only supporting up to 16 stops. We could support any number of stops, but because `Paint` has to hold the stops and is `Copy`, I had to impose a limit. If we removed `Copy` from `Paint` then we could put a `Vec` in and have any number of stops, but that would be a compatibility breaking public API change (and maybe not worth it? who needs more than 16 stops?).

iamralpht:
Ok, I redid this change to move the gradient storage to `Canvas` so new backends can more easily support multi-stop gradients. This also means that `Params` is unchanged and still mostly corresponds to the shader uniforms. I think this design is a much better.

I think this is the right way to go, but I'm also still learning about the underlying design of femtovg/nanovg, so it's entirely possible there's a better approach that I've missed. This change is ready to review now, and I'm going to work on some tests.

iamralpht:
I added a test (in the form of an example) which renders this on my computer:
gradient screenshot

One interesting thing to note in this example is that the "transparent red" in the two stop gradients turns into complete transparency. This is because the stop colors are premultiplied for the two-stop gradients, so the color information gets lost. In the multi-stop case we only do the premultiplication after we've evaluated the color in the gradient texture, so we don't lose anything. (We do currently turn a multi-stop gradient into a classic two-stop gradient if there are only two stops and they're positioned at 0.0 and 1.0, either we shouldn't do this or we should not premultiply two-stop gradient colors and do the multiplication in the shader).

src/gradient_store.rs
```diff
@@ -0,0 +1,150 @@
...
+
+use crate::{Color, ErrorKind, ImageFlags, ImageId, ImageInfo, ImageSource, Renderer, image::ImageStore, paint::MultiStopGradient};
+
+/// GradientStore holds image ids for multi-stop gradients. The actual image/textures
+/// are contained by the Canvas's ImageStore.
+//
+// If many gradients are used in a frame, we could combine them into a single texture
+// and update the texture immediately prior to giving the renderer the command list.
+pub(crate) struct GradientStore {
+    this_frame: BTreeMap<MultiStopGradient, ImageId>,
```
tronical, line 14: I’m just curious, is there a specific reason to prefer BTreeMap over HashMap here? Do you need the ordering of keys?

peter-todorov:
Wow, that's a really great addition to femtovg. Thank you, Ralph. I think there's a conflict with your other PR (a comma in the renderer).

src/gradient_store.rs
iamralpht: No, I used `BTreeMap` because it was easier to do `Ord` for the floating point values in `MultiStopGradient` than to do `Hash`. I bet if you have fewer than 5 multistop gradients in a frame that we'd actually be faster using a flat array and doing a linear search, but this was easier to implement and is unlikely to be very hot.

iamralpht:
@peter-todorov thanks to you and the other contributors for putting FemtoVG together; it has been a fun codebase to make changes to :)

src/gradient_store.rs
tronical: Makes perfect sense - thanks :)

PR #17, opened 2020-12-29
hide alpha mask functions from public API, and update README.

iamralpht:
The alpha mask can only really be used via draw_triangles which also isn't exported (and is only used when drawing text from an atlas). This commit changes the public API by removing Paint::set_alpha_mask and Paint::alpha_mask.

The README removes the note that only two stops are supported in gradients, because we now have multi-stop gradients.


peter-todorov:
Thank you Ralph! Merged.

PR #18, opened 2020-12-31
Fix WASM build

tronical:
Use the glow Framebuffer type alias for set_screen_target, which aliases
to Gluint for native GL and a different type for web_sys.

This also includes a small very basic CI setup. Sample build: https://github.com/tronical/femtovg/actions/runs/454885480

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,24 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


iamralpht:
Oops, sorry about that; I didn't try a wasm build, and for some reason thought it was always aliased to u32.

CI setup is really nice; hopefully we can get a software GL implementation running in CI too so we can do pixel tests.

adamnemecek:
@tronical Thanks!

PR #20, opened 2021-01-03
Make it possible to run the demo in the browser

tronical:
* canvas glue code in demo.rs takes care of fetching the canvas element.
* From the image crate we don't need the jpeg-rayon variant but the
  regular one, as rayon uses threads, which don't work out of the box
  in Rust's wasm32-unknown-unknown target.
* Assets like fonts and images are embedded into the binary
* The usvg crate has text disabled, because that would pull a bunch
  of dependencies that can't be cross-compiled.
* Use the instant crate to get a wasm working version of the
  performance counter.


adamnemecek:
Is there a reason why you are using winit 0.22 instead of 0.24? 0.22 doesn't run on Apple M1 laptops.

tronical:
It’s the one that the glutin version used depends on. I’d be happy to bump glutin and winit to a current version but that requires further minor unrelated changes due to API breakage. Would you prefer that to go into this MR or separate?

adamnemecek:
I think I would prefer it in this PR if that's not too much of a hassle.

tronical:
No problem at all, I’ll include it tomorrow.

tronical:
Fixed a typo in the resource deps, moved the jpeg dependency to dev-dependencies (the core crate shouldn't depend on a specific image format) and clarified the README a little.

PR #21, opened 2021-01-06
Add API to allow specifying the index for true type font collections

tronical:
This is particularly useful for example when using femtovg in
combination with font-kit, which returns results that point towards
system fonts that may be font collections and then the index is
relevant.

### DIFF
--- a/examples/breakout.rs
+++ b/examples/breakout.rs
@@ -91,13 +91,13 @@ impl Game {
 
         let fonts = Fonts {
             regular: canvas
-                .add_font("examples/assets/Roboto-Regular.ttf")
+                .add_font("examples/assets/Roboto-Regular.ttf", None)
                 .expect("Cannot add font"),
             bold: canvas
-                .add_font("examples/assets/Roboto-Bold.ttf")
+                .add_font("examples/assets/Roboto-Bold.ttf", None)
                 .expect("Cannot add font"),
             light: canvas
-                .add_font("examples/assets/Roboto-Light.ttf")
+                .add_font("examples/assets/Roboto-Light.ttf", None)
                 .expect("Cannot add font"),
         };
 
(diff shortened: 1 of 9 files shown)


adamnemecek:
Shouldn't the the index in Font::new optional as well?

tronical:
It could, but I’m curious: why? :) Font is not public API afaics.

That said, how do you feel about breaking API changes? I’m not so fond of the “overloads” and would rather make the index u32 a required parameter in the existing functions.

peter-todorov:
@tronical thank you for the PR. I think this is a good API addition and I like the suggestion to have an index: Option<u32> or just u32 instead of the "overloads". It will be easy and quick for people to update this in their code and we haven't yet made any stability guarantees.

tronical:
Great, I fully agree. Will make the changes.

tronical:
This is tempting, although I suppose it doesn’t work if you already have an Option?

The more I think about this API the more I dislike it, because if used with an actual TrueType collection it means the entire font data is copied N times.

So I’m contemplating researching the option of not doing what the current PR does but rather hide the entire font behind a trait, so that it’s possible to efficiently expose fonts without having to copy the data.

In the use case I have I mind I’d always want to mmap (or the ttf is embedded).

The same goes for efficient font fallback handling: I’d love to load fallback fonts lazily — the lists that macOS or font config return are rather long and it would be better to traverse them when glyphs cannot be found.

Anyway, I’d like to do some research here and hope to come back with a better API:)

adamnemecek:
It does work if you already have an option. Also you could make it Into<&T> thereby circumventing the thing.

tronical:
I'm cooking a better API that allows for proper font data sharing and index support in https://github.com/sixtyfpsui/femtovg/tree/simon/shared-fonts . Since GH doesn't allow changing the origin of a merge request, I'll close this one and will re-open a new one :)

PR #23, opened 2021-01-08
Add glyph margin outside the sampled area, keep padding inside. Fix #22

### LINKED ISSUE #22: Text rendering glitch, random noise around glyphs.
At first this seemed to affect only some machines, but it turns out to be a general problem.

Look closely at the black text on grey background and you’ll see random pixel lines around the glyphs.
The white-on-black text with red areas at the left side is a view of the font atlas texture.

Void Linux, AMD Radeon 5570:
femtovg-text-glitch-Radeon-5570 Captura de pantalla_2021-01-08_12-11-34

Raspbian, Raspberry Pi 400:
femtovg-text-glitch-RPi-400 Captura de pantalla_2021-01-08_10-34-40

Pop_OS (Ubuntu), Nvidia RTX 2080:
femtovg-text-glitch-RTX2080 Captura de pantalla 2021-01-08 09-31-16

The screenshots show a slightly modified femtovg, where each font atlas texture is filled-in with red when allocating it. That ensures that the glitch appears. In the default configuration, if the textures are allocated in fresh empty VRAM the glitch is not visible because the wrongly-sampled pixels happen to be transparent.

It does not appear in nanovg because they zero the texture memory when allocating each atlas texture.

The cause is a known source of confusion: when using bilinear filtering for rendering the textures, the color of a pixel depends on its neighbors, even if they lie outside the sampled area. See how, in this composite of two screenshots (one with bilinear filtering, another without) the red border, that I added around each glyph outside of the sampled area (the quad in the function `render_atlas`), encroaches into the black pixels of the glyph padding. That’s the glitch. Without that extra border, the interference pixels belong in many cases to the padding of their neighbors and have no visible effect. Then it only appears if the gaps between glyph boxes, which are uninitialized memory areas, contain values not close to zero.
bilinear-sampling

There is a brief explanation of the interaction between bilinear filtering and texel coordinates here: http://drilian.com/2008/11/25/understanding-half-pixel-and-half-texel-offsets/

## Solution
The solution is to add a 1-pixel *margin* of transparent pixels around the *padding* around the glyph box:
Captura de pantalla_2021-01-02_12-44-24 padding1margin1 diagram boxes

Looking at the font atlas texture, it seems that a padding of one pixel is enough, so changing `GLYPH_PADDING` from 2 to 1 allows us to add the margin without consuming more memory than before.

I’m finishing a PR with this fix.

AlbertoGP:


adamnemecek:
Thanks so much!

PR #26, opened 2021-01-12
Add font atlas debug functions, --features "debug_inspector"

AlbertoGP:
cargo run --example text --features "debug_inspector" --release
Debug builds additionally fill unused areas of the font textures
with red to highlight the difference with the glyph boxes.

Captura de pantalla_2021-01-12_19-25-27

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,7 @@ wasm-bindgen = { version = "0.2" }
 [features]
 default = ["image-loading"]
 image-loading = ["image"]
+debug_inspector = []
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


peter-todorov:
@AlbertoGP thank you for the PR. I think `debug_inspector_get_textures` it's a really good idea to have as a canvas API. I even feel like that shouldn't be future gated, as it may be useful in other cases as well.

AlbertoGP:
My reason for having it feature-gated is that using a texture atlas for text rendering is an implementation detail, and thus should not be exposed in the stable API.

PR #30, opened 2021-02-11
Improve text rendering

ollpu:
This PR implements three improvements / fixes to text rendering.

(open the images up to 100% scale to see the difference properly)

Before:
text example before

After:
text example after

Side-by-side:
image

### Vertical pixel alignment

When copying from the atlas, glyphs were placed with subpixel offsets in the y-direction and therefore bilinear interpolation was always used. This caused smearing in the vertical direction.

Text is now rendered into the atlas with its baseline aligned to an integer coordinate and glyphs are only copied to integer y-coordinates. I also made `FontMetrics::height` rounded in order to make line spacing consistent.

Follow-up: Should we offer a non-quantized mode for text separately? This would allow smoother scaling animations or finer spacing when needed. I think that mode should always render directly or use a fixed number of oversampled atlas scales. Or do the same subpixel quantization as with x?

### Horizontal subpixel quantization

The quantization applied to glyph x-coordinates only affected the cache ID, but not the actual rendering itself. This should make results more consistent (at most 1/20 px error instead of 1/10).

### Anti-aliasing rook pattern

The anti-alias filter pattern seems to have had a typo, since it looked like this when plotted:
image
This is now corrected, subtly improving quality.


peter-todorov:
Awesome work ollpu! Thank you for the contribution! Those are really great improvements I can't tell you how much time I spent trying to fix the blurriness.

> Follow-up: Should we offer a non-quantized mode for text separately?

Initially, text rendering was direct, but it had two problems:
- The nanovg AA algorithm degrades on small sizes.
- It was much slower than using an atlas.

I tried disabling the nanovg AA and using that rook pattern to render the glyphs, but that just made the performance issue worse (8 times worse). And as much as I tried I couldn't improve the nanovg AA for small sizes. 
So I decided to try with a multisampled framebuffer for the atlas. Initially, the results were quite good (even at 16x) and the GPU did a faster job of it. But that introduced **a lot** of complications since now the atlas framebuffer needed to be resolved to a texture, before being used, after every new glyph inserted. It also had a bunch of edge cases, bugs, and driver issues (I think it didn't work at all on osx).

In the end, I went with the current solution, which is not that great.

You've probably already seen this, but here's how direct rendering looks (change the threshold value to 1 or something):
https://github.com/femtovg/femtovg/blob/9ba9422be4e7543984b3d2b755e09d8b3d6bd7dc/src/lib.rs#L1085-L1087

Result:
DeepinScreenshot_xfdesktop_20210212111940

Sidenote: If you change the fringe offset in `path/cache.rs` to `0` it almost fixes this glitch by making everything ticker. I haven't found a way yet to make it render correctly for small font sizes. Also compiling in release improves the speed a bit:
https://github.com/femtovg/femtovg/blob/9ba9422be4e7543984b3d2b755e09d8b3d6bd7dc/src/path/cache.rs#L560

DeepinScreenshot_xfdesktop_20210212112415

At this point even if the nanovg AA can be made to work for small sizes it may be too slow to use for some people. The best solution would be if we just use nanovg AA for everything and shift-delete the whole texture/caching stuff, but I'm not sure if it can be optimized to ~1ms.

The oversampled atlas scales you suggested may give the best quality/performance tradeoff at this point.

Again thank you for the contribution!

ollpu:
@peter-todorov Interesting!

I believe [@geom3trik](https://github.com/geom3trik) tried the oversampled atlas to some degree of success.

PR #32, opened 2021-03-06
Added with_-constructors to Paint to simplify setting of parameters

porky11:
Suggestion: Added constuctors of the form `with_` to `Paint`, which simplifies creation of specific paints.

### DIFF
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -492,6 +492,12 @@ impl Paint {
         self.flavor = PaintFlavor::Color(color);
     }
 
+    /// Returns the paint with a new solid color set to the specified value.
+    pub fn with_color(mut self, color: Color) -> Self {
+        self.set_color(color);
+        self
+    }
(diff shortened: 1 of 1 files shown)


tronical:
Nice, this also avoids having to use a mutable local paint variable, right? So

```rust
let mut paint = Paint::color(...);
paint.set_line_width(...);
```

can be just written as

```rust
let paint = Paint::color(...).with_line_width(...);
```

Is that correct?

porky11:
@tronical Right, it should work that way.
Didn't really use it yet myself, though, since I try to stick to the published versions.

Lemmih:
I think it would be worthwhile to have similar methods on `Path` (`with_move_to`, `with_line_to`, etc) and on `Transform2D` (`translated`, `scaled`, etc, keeping with the `inverse`/`inversed` naming scheme).

PR #38, opened 2021-04-25
Fix polygon area calculation

adroitwhiz:
The previous algorithm was incorrect because it didn't consider the triangle of the last two points and the first point, or of the last point and the first two points. It's been replaced with [one that works without calculating triangle area at all](https://stackoverflow.com/a/1165943).

This fixes some shapes being "randomly" thickened by the AA algorithm:
Before | After
--- | ---
image | image

Holes in glyphs are still incorrectly thickened, but that's because they're clockwise shapes using the even-odd fill rule, so there's no way to tell that they're holes, and they get offset inwards. That seems to be a much harder problem to fix.

### DIFF
--- a/src/geometry.rs
+++ b/src/geometry.rs
@@ -13,15 +13,6 @@ pub fn quantize(a: f32, d: f32) -> f32 {
     (a / d + 0.5).trunc() * d
 }
 
-pub fn triarea2(ax: f32, ay: f32, bx: f32, by: f32, cx: f32, cy: f32) -> f32 {
-    let abx = bx - ax;
-    let aby = by - ay;
-    let acx = cx - ax;
-    let acy = cy - ay;
-
-    acx * aby - abx * acy
-}
-
 pub fn pt_equals(x1: f32, y1: f32, x2: f32, y2: f32, tol: f32) -> bool {
     let dx = x2 - x1;
     let dy = y2 - y1;
--- a/src/path/cache.rs
+++ b/src/path/cache.rs
@@ -113,12 +113,8 @@ impl Contour {
     fn polygon_area(points: &[Point]) -> f32 {
         let mut area = 0.0;
 
-        for window in points.windows(3) {
-            let p0 = window[0];
-            let p1 = window[1];
-            let p2 = window[2];
-
-            area += geometry::triarea2(p0.x, p0.y, p1.x, p1.y, p2.x, p2.y);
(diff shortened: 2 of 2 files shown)


adroitwhiz:
It looks like I can get antialiasing looking just about perfect for text with a couple hacks:
image

The hacks are:
- Render text with the non-zero winding fill rule (which is what TrueType uses anyway), and unconditionally reverse path point order rather than only reversing them for paths with a solidity of `Solidity::Hole`. This is a hack because there's no (fast, general) way to tell which subpaths are holes--you could, for instance, have a clockwise hole inside a counterclockwise path, and you'd have no way of knowing.
- Always render only the outer half of the antialiasing fringe. This is a hack because it means that e.g. the inside of the even-odd star path will not be antialiased.

The hard part is generalizing these hacks to work over all shapes.

tronical:
I can’t speak to the code changes but I love to see improvements to the AA quality for text!

ollpu:
@adroitwhiz Something that might be affecting your results is that you seem to be using DPI scaling (judging by the incorrectly drawn RUST-texts :D). Fringe width is currently set incorrectly when the scale isn't 1, among other issues.

adroitwhiz:
> @adroitwhiz Something that might be affecting your results is that you seem to be using DPI scaling (judging by the incorrectly drawn RUST-texts :D). Fringe width is currently set incorrectly when the scale isn't 1, among other issues.

The DPI is 1--I'm not sure why the gradients are drawn incorrectly.

peter-todorov:
I've been searching a lot for this issue and I never suspected the polygon area calculation :facepalm: I assumed it is correct from nanovg. 

Also, the font rendering hack looks really good, if we could use the native AA in the text rendering path and not that 8x offset hack it would speed up glyph generation significantly. The issue with the gradients in the RUST text is that glyphs are scaled by the em square size (2048?) and that scaling is incorrectly applied to the gradient in the paint as well.

ollpu:
Just to check, nanovg was doing it a bit differently, but it seems correct:

```c
static float nvg__polyArea(NVGpoint* pts, int npts)
{
	int i;
	float area = 0;
	for (i = 2; i < npts; i++) {
		NVGpoint* a = &pts[0];
		NVGpoint* b = &pts[i-1];
		NVGpoint* c = &pts[i];
		area += nvg__triarea2(a->x,a->y, b->x,b->y, c->x,c->y);
	}
	return area * 0.5f;
}
```

PR #40, opened 2021-05-19
Add support for applying a gaussian blur filter to images

tronical:
This adds the

```rust
   pub fn filter_image(&mut self, target_image: ImageId, filter: ImageFilter, source_image: ImageId)
```

function to the renderer, where the ImageFilter is an enum that starts
with a gaussian blur and the standard deviation parameter.

The GL renderer renders this in two passes, a horizontal and a vertical
blur.

Apart from adding filter code path, this patch makes one other change to
the rendering interface: The drawing commands are now moved from the
canvas to the renderer. This is needed because then the second pass in
the blur filter can just change the Command's image parameter, as it
consumes it. Otherwise Command would have to implement Clone, which
would likely be expensive due to the Vec fields within.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "femtovg"
 description = "Antialiased 2D vector drawing library"
-version = "0.1.3"
+version = "0.2.0"
 license = "MIT/Apache-2.0"
 readme = "README.md"
 authors = ["Peter Todorov <ptodorov@cytec.bg>", "Adam Nemecek <adamnemecek@gmail.com>"]
--- a/examples/paint_filter.rs
+++ b/examples/paint_filter.rs
@@ -1,0 +1,140 @@
+/**
+ * Shows how to use Canvas::filter_image() to apply a blur filter.
+ */
+use instant::Instant;
+
+use resource::resource;
+
+use glutin::{
+    event::{
+        Event,
+        WindowEvent,
+    },
+    event_loop::{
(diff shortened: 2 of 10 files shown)


tronical:
For some context, we're trying to use this to implement a better looking drop-shadow in SixtyFPS - by applying the HTML Canvas algorithm that blurs the shadow of the element to be rendered on top - a potentially round rectangle in our case.

https://github.com/sixtyfpsui/sixtyfps/pull/229

tronical:
@adamnemecek @peter-todorov what's your take on this? :)

adamnemecek:
Looks good I guess? Haha, I’m not sure.

tronical:
Hehe. There are three aspects to consider perhaps:

(1) is the api okay, flexible for future changes?
(2) is the way the change integrates with the existing rendering code and backend api (change!) okay? Maybe I overlooked some caveats or made wrong assumptions?
(3) Is the implementation of the filter itself okay? There’s room for improvement, in terms of performance for example. I think we have some freedom there.

I’d love to get approval for this change and then make a new 0.2.0 release, so that we can make use of this in our next release :-)

peter-todorov:
Thanks @tronical ! It looks pretty good to me. I had a version of this a while back, but it was with a second shader that induced a lot of pipeline switches and I ran the horizontal and vertical blur multiple times using ping-pong textures until I got to the desired blur level. The approach here is much better. Thanks again for your contribution!

PR #41, opened 2021-05-26
Expose `TextContext` as "headless" API to measure text and share fonts across canvas instances

tronical:
The API is basically the same as what's already expose in Canvas, with the difference that `TextContext` is using interior mutability and explicit sharing behind the scenes.


tronical:
Thank you!

PR #57, opened 2021-10-19
Require rustfmt from +stable rather than +nightly.

Lemmih:


Lemmih:
If this PR is acceptable then I would prefer it to be squashed.

PR #59, opened 2021-10-19
Set Minimum Supported Rust Version (MSRV) to 1.46

Lemmih:


tronical:
I'm in favour of defining and testing this. Looks like it'll have to be a newer version though :)

> warning: slotmap requires rustc => 1.49.0

tronical (approved):

PR #66, opened 2021-10-26
Provide a safe way to initialize an OpenGl renderer via glutin.

Lemmih:
`OpenGl::new` is quite low-level and will segfault if used incorrectly. This PR introduces the `OpenGl::new_glutin` method which will safely produce an OpenGl renderer from a windowed context.

This PR also marks `OpenGl::new` as unsafe which completely breaks backwards compatibility. Maybe we should mark `OpenGl::new` as deprecated instead?


tronical (approved):
While I think most feedback was in favour of breaking the API, this change is safe and good in any case. I think it should go in. The breakage can be a separate change along with the version bump at a later point. I'm in favour of submitting this @Lemmih  :)

PR #67, opened 2021-11-01
Created a position type

porky11:
The new position type is a simple vector type, which only implements the most important operations.

This type is only used internally. The public API stays the same.

It's mostly used to simplify cases, where the same calculation is repeated for x and y.

To see, how the code is simplified, see commit b7a9bf6. You need a little extra knowledge about the position type, but it makes the code more compact and less redundant.
I also think, it's clearer what is done internally this way, for example when the normal of the vector is taken (see `orthogonal`).

If this should be merged, we might change some names:
* `Position` could also be called `Position2D`, `Vector` or `Vector2`
* `orthogonal` could also be called `normal`
* should `dot`, `angle` or `mirror` be dropped, since they are not used often?

### DIFF
--- a/src/path/cache.rs
+++ b/src/path/cache.rs
@@ -4,7 +4,7 @@ use std::ops::Range;
 
 use bitflags::bitflags;
 
-use crate::geometry::{self, Bounds, Transform2D};
+use crate::geometry::{Bounds, Transform2D};
 use crate::position::Position;
 use crate::renderer::Vertex;
 use crate::utils::VecRetainMut;
@@ -253,9 +253,8 @@ impl PathCache {
                     points.get_mut(i - 1).unwrap()
                 };
 
-                p0.dpos.x = p1.pos.x - p0.pos.x;
-                p0.dpos.y = p1.pos.y - p0.pos.y;
-                p0.len = geometry::normalize(&mut p0.dpos.x, &mut p0.dpos.y);
+                p0.dpos = p1.pos - p0.pos;
+                p0.len = p0.dpos.normalize();
 
                 bounds.minx = bounds.minx.min(p0.pos.x);
                 bounds.miny = bounds.miny.min(p0.pos.y);
--- a/src/path/cache.rs
+++ b/src/path/cache.rs
@@ -41,12 +41,12 @@ impl Point {
     }
 
     pub fn is_left(p0: &Self, p1: &Self, x: f32, y: f32) -> f32 {
-        (p1.pos.x - p0.pos.x) * (y - p0.pos.y) - (x - p0.pos.x) * (p1.pos.y - p0.pos.y)
+        let pos = Position { x, y };
+        (p1.pos - p0.pos).dot((pos - p0.pos).orthogonal())
     }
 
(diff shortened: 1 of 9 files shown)


tronical:
I love the way this simplifies the internals. I'm wondering though: Wouldn't it be easier to re-use euclid (what everyone else is using)?

porky11:
We already had a small discussion about that on Discord. That's what I wrote:

> I don't think, switching to euclid would improve anything. It might just be bloat.
> The new type I added only implements addidion, subtraction, scalar multiplication and only a few special methods, which are needed.
> 
> What I don't like about euclid:
> * it's both, 2D and 3D; I'd prefer if they would split it into a 2D and a 3D version
> * there's a difference between points and vectors, which sounds nice in theory, but in practice, you'll probably end up converting back and forth between them all the time
> * we basically only need some vector type, not all the other types; if we can at least use more than that, maybe the Transform2D, it would be more reasonable
> 

Besides that, the euclid vector and point types don't seem to implement one of the methods, which I added to my specific implementation, since it's useful for femtovg: `orthogonal`, which creates an orthogonal vector of the same length.

tronical:
> It might just be bloat.

Why bloat? I see opportunity to share (generated) code when applications are also using euclid (likely, and even more if we decide to utilise it in the API).

porky11:
I don't like the idea of exposing a specific math library in an API.

I'd rather expose tuples, vectors or [mint](https://docs.rs/mint/) types.

Or make the API generic using some `Into`- and `From`-traits.

I agree about sharing code, but I would prefer to use a more minimal vector library:

* only vectors, no points
* only 2D (or dimension generic using const generics)
* maybe a few less types

(basically the points I already made)

But since no such library exists, nor is established, using euclid might be the best solution for now.

porky11:
I just tried to convert it to euclid, but the vector types are pretty stupid: They have an additional private field.
So I can't just do this:

```rust
fn test(x: f32, y: f32) {
    let pos = Vector { x, y };
    let Vector { x, y } = pos;
}
```

I can't

porky11:
A different pull request based on this one for a euclid version #68

Lemmih:
I'm thinking we merge this PR first and then think about `euclid` at a later point in time.

src/path/cache.rs
```diff
@@ -23,33 +24,29 @@ bitflags! {
 
 #[derive(Copy, Clone, Debug, Default)]
 pub struct Point {
-    x: f32,
-    y: f32,
-    dx: f32,
-    dy: f32,
+    pos: Position,
+    dpos: Position,
```
Lemmih, line 28: `dx`/`dy` is a vector, not a position.

porky11: I guess, I should just rename Position to Vector

porky11: Positions are not vectors, but a vector can always be used to represent a position.

Lemmih: I would greatly prefer having both positions and vectors. Otherwise it is not clear what the data represents.

porky11:
I created a new pull request #72

PR #69, opened 2021-11-03
Add support for device-pixel-ratio in examples.

Lemmih:
This PR also fixes a bug in `measure_font` which modified the result using the current transformation.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -50,7 +50,7 @@ fn main() {
         use glutin::ContextBuilder;
 
         let wb = WindowBuilder::new()
-            .with_inner_size(winit::dpi::PhysicalSize::new(1000, 600))
+            .with_inner_size(winit::dpi::LogicalSize::new(1000, 600))
             .with_title("femtovg demo");
 
         //let windowed_context = ContextBuilder::new().with_gl(GlRequest::Specific(Api::OpenGlEs, (2, 0))).with_vsync(false).build_windowed(wb, &el).unwrap();
@@ -156,6 +156,13 @@ fn main() {
 
     let mut perf = PerfGraph::new();
 
+    {
+        #[cfg(not(target_arch = "wasm32"))]
+        let window = windowed_context.window();
+        let dpi_factor = window.scale_factor();
+        canvas.scale(dpi_factor as f32, dpi_factor as f32);
+    }
+
     el.run(move |event, _, control_flow| {
         #[cfg(not(target_arch = "wasm32"))]
         let window = windowed_context.window();
@@ -194,7 +201,14 @@ fn main() {
                         canvas.scale(1.0 + (y / 10.0), 1.0 + (y / 10.0));
                         canvas.translate(-pt.0, -pt.1);
                     }
-                    _ => (),
+                    winit::event::MouseScrollDelta::PixelDelta(pos) => {
+                        let y = pos.y as f32;
+                        let pt = canvas.transform().inversed().transform_point(mousex, mousey);
+                        let rate = 2000.0;
+                        canvas.translate(pt.0, pt.1);
+                        canvas.scale(1.0 + (y / rate), 1.0 + (y / rate));
+                        canvas.translate(-pt.0, -pt.1);
+                    }
                 },
                 WindowEvent::MouseInput {
                     button: MouseButton::Left,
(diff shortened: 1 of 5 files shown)


archseer:
@tronical Could you take a look at this PR? It seems like a reasonable improvement

tronical:
> @tronical Could you take a look at this PR? It seems like a reasonable improvement

Ah yes. Thanks for the ping. Generally this looks like a good idea to me.

IMO a nicer way would be to start using Euclid types again for this.

I’ll take a closer look on Monday, added to my todo.

examples/svg.rs
```diff
@@ -52,6 +52,13 @@ fn main() {
 
     println!("Path mem usage: {}kb", total_sisze_bytes / 1024);
 
+    {
+        #[cfg(not(target_arch = "wasm32"))]
+        let window = windowed_context.window();
+        let dpi_factor = window.scale_factor();
+        canvas.scale(dpi_factor as f32, dpi_factor as f32);
```
tronical, line 59: Why is this guarded against the wasm build? Shouldn't it be used there, too?

examples/svg.rs
```diff
@@ -154,6 +168,7 @@ fn main() {
 
                 canvas.save();
                 canvas.reset();
+                canvas.scale(dpi_factor as f32, dpi_factor as f32);
```
tronical, line 171: Why is this scaling with the scale factor what looks like a second time?

src/lib.rs
```diff
@@ -251,6 +251,14 @@ struct State {
     alpha: f32,
 }
 
+impl State {
+    fn with_pixel_ratio(pixel_ratio: f32) -> Self {
```
tronical, line 255: Perhaps the `Default` impl should be removed for `State` to make sure to catch all cases where `State` is re-initialized and a decision should be made whether to apply the pixel ratio or not. It could be `new()` and `new_with_pixel_ratio` perhaps?

src/lib.rs
```diff
@@ -345,12 +353,9 @@ where
     pub fn set_size(&mut self, width: u32, height: u32, dpi: f32) {
         self.width = width;
         self.height = height;
-        self.fringe_width = 1.0 / dpi;
-        self.tess_tol = 0.25 / dpi;
-        self.dist_tol = 0.01 / dpi;
         self.device_px_ratio = dpi;
 
-        self.renderer.set_size(width, height, dpi);
+        self.renderer.set_size(width, height);
```
tronical, line 358: Good move to remove this from the renderer :)

examples/text.rs
```diff
@@ -33,6 +33,7 @@ fn main() {
         window_size.height as u32,
         windowed_context.window().scale_factor() as f32,
     );
+    canvas.reset();
```
tronical, line 36: Do you happen to remember why this call to `reset()` is needed here?

src/lib.rs
```diff
@@ -384,6 +389,16 @@ where
...
 
+    /// Returns the width of the current render target.
+    pub fn logical_width(&self) -> f32 {
+        self.width() / self.device_px_ratio
+    }
+
+    /// Returns the height of the current render target.
+    pub fn logical_height(&self) -> f32 {
+        self.height() / self.device_px_ratio
+    }
```
tronical, line 400: I think the documentation should explain the difference between this and the physical size getters.

I'd even go as far as even renaming `width()` and `height()` to `physical_width()` and `physical_height()`. This requires a major version bump, of course - but why not :-)

src/text.rs
```diff
@@ -901,7 +901,7 @@ impl GlyphAtlas {
 
         // render glyph to image
         canvas.save();
-        canvas.reset();
+        canvas.hard_reset();
```
tronical, line 904: This makes me wonder if perhaps `set_render_target` should be changed. I mean, won't pretty much **any** user of `set_render_target` when rendering into an image want to call `hard_reset()` to reset the dpi otherwise?

tronical (commented):
Nice!

I think this leaves a few questions about the API and documentation, but it makes generally sense to me.

examples/demo.rs
```diff
@@ -156,6 +156,14 @@ fn main() {
 
     let mut perf = PerfGraph::new();
 
+    {
+        #[cfg(not(target_arch = "wasm32"))]
+        let window = windowed_context.window();
+        let dpi_factor = window.scale_factor();
+        canvas.set_size(0, 0, dpi_factor as f32);
+        canvas.reset();
+    }
```
archseer, line 165: Note: I had to copy this under the `Resized` event for scale to work correctly on resize

PR #73, opened 2021-11-06
Vector API [WIP]

porky11:
I started changing the API.
Currently only the methods of path don't take `(x: f32, y: f32, ...)` anymore, but instead `<T: Into<[f32; 2]>>(pos: T, ...)`.

This way, you could use femtovg in basically two ways:

1:
A more traditional approach, but with extra square brackets (`path.move_to([x, y])` instead of `path.move_to(x, y)`).
In this case, it's a bit more syntax, but I think, it's also worth it because it gets clearer, which parameters belong together.
And even auto formatting packs parameters, which belong together in the same line, when multiple points are supplied.

2:
Supply vectors of an arbitrary vector math library in a more straightforward way (`path.move_to(pos + offset)` instead of `let start_pos = pos + offset; path.move_to(pos.x, pos.y)`).

Alternatives, I decided against:

1. use (mint)[https://docs.rs/mint] -> problem: [f32; 2] does not implement `IntoMint`, so the more traditional approach won't work
2. use two versions of each method (one for vectors, one for parameter pairs) -> sounds like bloat, but this way, the change would not be a breaking change at least
3. use some specific vector library -> same problems as alternative 1, but not as easy to use with other vector libraries
4. same approach but with tuples (`(f32, f32)`) instead of arrays (`[f32; 2]`) (why are arrays not just another way to express tuples of the same element type?)

TODO:
- [ ] convert w/h pairs and radius pairs to a `impl Into<[f32; 2]>`
- [ ] also convert x/y or w/h pairs of other modules (most important paint) 
- [ ] also store [f32; 2] for x/y pairs in public structs (like `Vertex`, already done for `TextMetrics`)
- [ ] make non-generic private versions of the methods and inline the generic ones, so only one version of each method gets compiled (or is the compiler already smart enough?)

Do you think, this change is worth it?
It will be a breaking change, and upgrading will cause a lot of work. But since most people use a vector library anyway, the resulting code will probably be better.


src/path.rs
```diff
@@ -136,38 +142,34 @@ impl Path {
     // Path funcs
 
     /// Starts new sub-path with specified point as first point.
-    pub fn move_to(&mut self, x: f32, y: f32) {
-        self.append(&[PackedVerb::MoveTo], &[x, y]);
+    pub fn move_to<T: Into<[f32; 2]>>(&mut self, pos: T) {
```
Lemmih, line 145: I'd prefer: `pub fn move_to(&mut self, pos: impl Into<[f32; 2]>) {`

Lemmih:
While this API is more convenient in some cases, I do think there is a lot of value in an API that's as simple as possible.

src/path.rs
porky11: I also got this idea after I pushed.

porky11:
> While this API is more convenient in some cases, I do think there is a lot of value in an API that's as simple as possible.

I agree. I also like, that you can just supply values and don't have to use vectors.

That's the main reason, why I used inbuilt array types. You can use it almost as before and no fancy types are introduced.
I'd argue this way it's not necessarily less convenient. You need a few more brackets, but also get more structure.

Some other alternative, I didn't mention, is just not using `impl Into<[f32; 2]>` but instead using just `[f32; 2]`.
The API would stay simpler, while using vectors is also less cumbersome (`path.move_to((pos + offset).into())` instead of `let pos = pos + offset; path.move_to(pos.x, pos.y)`).

PR #76, opened 2022-01-18
Allow tinting image paints

rhelmot:
The shader has logic in it for using a full tint instead of just an alpha tint, however, it was not exposed to the user. This adds an alternate image_tint constructor which allows providing a tint color instead of an alpha.


tronical:
I like that :)

peter-todorov:
I also think that's a good addition :+1:

peter-todorov (approved):

PR #77, opened 2022-01-22
Reserve space in vectors in fill_path and stroke_path

sky1e:
I was profiling an app I'm working on and noticed a noticeable amount of time was being spent in `fill_path` reallocating vectors.
flamegraph before 21.5% of samples in this app were in `femtovg::Canvas<T>::fill_path`.

So I added lines to reserve sufficient space in the vectors ahead of inserting elements
flamegraph after Now only 15.9% of samples are in `femtovg::Canvas<T>::fill_path`.


tronical (approved):
👍

PR #81, opened 2022-02-18
Create femtovg book for new users

rMazeiks:
If I'm not mistaken, the project's missing an introductory guide for new users. Examples are nice, but it takes a while to actually figure out which bits are the most important ones.

Lemme know what you think! This is obviously not finished but already provides some value.

This is an [`mdbook`](https://rust-lang.github.io/mdBook/) book and can be run with `mdbook serve book --open`.


tronical:
That’s a great idea!

One suggestion I have is to integrate the mdbook build into the CI. It might also make sense to use external files for some of the examples so that their build can be tested. We did something like that in Slint with mdbook and `{{include}}`: https://github.com/slint-ui/slint/blob/master/docs/tutorial/rust/src/creating_the_tiles_from_rust.md

tronical (approved):
I think this provides already value, so r=me from my side :)

rMazeiks:
> use external files for some of the examples so that their build can be tested

Great idea, done! Now, the final code in each sub-chapter is an example, which automatically gets compiled during `cargo test`.

> One suggestion I have is to integrate the mdbook build into the CI

Agreed, it would be good to build & host this somewhere automatically. Perhaps on [femtovg.github.io](https://github.com/femtovg/femtovg.github.io), which currently has nothing? Anyway, I guess this would have to be done by someone who know stuff about CI :)

tronical:
> > use external files for some of the examples so that their build can be tested
> 
> Great idea, done! Now, the final code in each sub-chapter is an example, which automatically gets compiled during `cargo test`.

Thanks.

> > One suggestion I have is to integrate the mdbook build into the CI
> 
> Agreed, it would be good to build & host this somewhere automatically. Perhaps on [femtovg.github.io](https://github.com/femtovg/femtovg.github.io), which currently has nothing? Anyway, I guess this would have to be done by someone who knows stuff about CI :)

Yes, femtovg.github.io is a good idea.

tronical:
I'll try to approve the CI runs, but you can also try this in a fork of yours first if you'd like faster turnarounds :)

tronical:
Might need a `cargo fmt` to silence the format check of the CI :)

rMazeiks:
Yeah, I'm currently testing the action by pushing to my own repo. I'll update the repo to the femtovg/ one once its successful.

I'll keep you updated! Sorry for the CI noise, yea i'll fix the formatting too

tronical:
> Sorry for the CI noise, yea i'll fix the formatting too

No worries - doesn’t bother me at all :).

rMazeiks:
yay! looks like the deploy action worked (see [here](https://github.com/rMazeiks/femtovg.github.io))

I've updated the target repository to femtovg/femtovg.github.io, so it should be good to merge.

However, since it wants to deploy to another repository, you'll need to set up an SSH deploy key ([see instructions here](https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-create-ssh-deploy-key)) In short, this repo needs the `ACTIONS_DEPLOY_KEY` secret set to the private key, and femtovg/femtovg.github.io needs a deploy key with "Allow write access" and the corresponding public key.

tronical:
👍
@adamnemecek can you help with the femtovg.github.io setup? (I don’t have access to they repo)

porky11:
No need for stuff like this.
Having simple examples which are maintained is already much better than most libraries.

Besides that, there's the [documentation](https://docs.rs/femtovg)

rMazeiks:
@porky11 ok

I found the examples too complicated and the documentation hard to get started with. I wanted something with a bit more hand holding. So I wrote this book I wish I had when I started learning femtovg.

I don't mind though, feel free to close if you don't think this is needed. Though I would recommend adding some more simple examples in that case (e.g. a simple window+rectangle+event loop, feel free to take my 80-line example from this PR). The existing examples are 500+ lines long and hard to follow.

geom3trik:
I would disagree here, I think a getting started guide like this is very useful for brand new users and at worst I see no downside to having one. I'm actually curious what's blocking this, is it the femtovg.github.io setup? @tronical

tronical:
> I would disagree here, I think a getting started guide like this is very useful for brand new users and at worst I see no downside to having one. I'm actually curious what's blocking this, is it the femtovg.github.io setup? @tronical

I agree, I also see no downside. Access to the repo is blocking this afaik.

EarthenSky (commented):
as a new user this tutorial was helpful

peterprototypes:
I know it has been more than a year, but I'm finally having time again for open source. This looks great and a book was always on my list of tasks for femtovg. @rMazeiks thank you for the PR

peterprototypes:
Done https://femtovg.github.io/

PR #82, opened 2022-02-19
Fix a compile error on Linux

sumibi-yakitori:
Currently, if I try to use this crate on Linux, I will get a compilation error.

[The case of running the examples does not have this problem.](https://github.com/femtovg/femtovg/blob/c0b7c47606235da8ff96033c6a5ba4a7566ddbf1/Cargo.toml#L62)

```sh
error: Please select a feature to build for unix: `x11`, `wayland`
  --> /home/yakitori/.cargo/registry/src/github.com-1ecc6299db9ec823/winit-0.26.1/src/platform_impl/linux/mod.rs:10:1
   |
10 | compile_error!("Please select a feature to build for unix: `x11`, `wayland`");
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This PR will fix it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ ouroboros = { version = "0.14" }
 
 [target.'cfg(not(target_arch = "wasm32"))'.dependencies]
 glow = { version = "0.11.0", default-features = false }
-glutin = { version = "0.28.0", optional = true, default-features = false }
+glutin = { version = "0.28.0", optional = true }
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 glow = { version = "0.11.0", default-features = false }


tronical:
Hmm, it’s true that a “plain” build of just the crate requires more than “cargo build” but rather for example `cargo build —features glutin/x11` (or so). But in any say application crate that’s using femtovg and that enables femtovg’s glutin feature, this application crate must also either disable the glutin feature if it isn’t using glutin or it must select which glutin backends it wants (or enable all of them via default).

How did you run into this build issue?

sumibi-yakitori:
Oh, I see.
So I should not use the `glutin` feature if I don't need to? Thanks.

I ran into this problem using this crate from `gtk4-rs`.
In this case, `glutin` is certainly not needed.

Not related to this issue, but when I try to use it from `gtk4`, it renders well on macOS, but seems to cause OpenGL errors on Linux.

tronical:
> Oh, I see. So I should not use the `glutin` feature if I don't need to? Thanks.

yep :)

> I ran into this problem using this crate from `gtk4-rs`. In this case, `glutin` is certainly not needed.
> 
> ```rust
> let renderer = OpenGl::new_from_function(epoxy::get_proc_addr)?;
> let canvas = Canvas::new(renderer)?;
> ```
> 
> Not related to this issue, but when I try to use it from `gtk4`, it renders well on macOS, but seems to cause OpenGL errors on Linux.

Hm, this could be something related to the kind of GL context created. Do you know what kind of profile or context it is? If you have a sample repo somewhere I can also have a look. Perhaps there is some detection missing in femtovg.

sumibi-yakitori:
Thanks for your suggestion.
I have created [a repository](https://github.com/sumibi-yakitori/femtovg-on-gtk4) for minimal problem reproduction.

It may not be a femtovg problem, it may be a GTK4 problem, or a graphics driver problem.

tronical:
I haven’t tried it myself yet (not on laptop at the moment), but glancing at the code and docs makes me wonder: perhaps you need to make the gtk::GLArea‘s context current at the beginning of the realize function. The docs have an example that suggests that this may be necessary: https://docs.gtk.org/gtk4/class.GLArea.html#drawing-with-gtkglarea

I’ve seen effects like these before that on some windowing systems it is not necessary to call make_current, for example because the context was never made not current or there exists only one, while on others it is strictly necessary to stick to the “rules” :)

sumibi-yakitori:
You were right.
By running `make_current` it rendered fine.
I really appreciate it!

tronical:
👍
I think then we can close this merge request as well :)

(please don’t hesitate to re-open if I missed something)

PR #84, opened 2022-03-03
Add methods on TextContext to resize caches

### LINKED ISSUE #83: Make shaping LRU caches tunable
I'm developing a GUI application which needs to layout a couple thousand lines of text on each refresh. It would be nice if I could nudge the `LRU_CACHE_CAPACITY` upwards a little bit, just for my application.

rhelmot:
Closes #83

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -270,14 +270,14 @@ impl TextContext {
 
     /// Adjusts the capacity of the shaping run cache. This is a cache for measurements of whole
     /// strings.
-    pub fn resize_shaping_run_cache(&mut self, capacity: usize) {
+    pub fn resize_shaping_run_cache(&self, capacity: usize) {
         self.0.as_ref().borrow_mut().resize_shaping_run_cache(capacity)
     }
 
     /// Adjusts the capacity of the shaped words cache. This is a cache for measurements of
     /// individual words. Words are separated by
     /// [UAX#29 word boundaries](http://www.unicode.org/reports/tr29/#Word_Boundaries).
-    pub fn resize_shaped_words_cache(&mut self, capacity: usize) {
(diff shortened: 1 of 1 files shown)


tronical:
There seems to be a generic CI issue at GitHub, so that failure is unrelated to this patch AFAICS.

src/text.rs
```diff
@@ -267,6 +267,19 @@ impl TextContext {
     pub fn measure_font(&self, paint: Paint) -> Result<FontMetrics, ErrorKind> {
         self.0.as_ref().borrow_mut().measure_font(paint)
     }
+
+    /// Adjusts the capacity of the shaping run cache. This is a cache for measurements of whole
+    /// strings.
+    pub fn resize_shaping_run_cache(&mut self, capacity: usize) {
```
tronical, line 273:
```suggestion
    pub fn resize_shaping_run_cache(&self, capacity: usize) {
```

src/text.rs
```diff
@@ -267,6 +267,19 @@ impl TextContext {
...
+        self.0.as_ref().borrow_mut().resize_shaping_run_cache(capacity)
+    }
+
+    /// Adjusts the capacity of the shaped words cache. This is a cache for measurements of
+    /// individual words. Words are separated by
+    /// [UAX#29 word boundaries](http://www.unicode.org/reports/tr29/#Word_Boundaries).
+    pub fn resize_shaped_words_cache(&mut self, capacity: usize) {
```
tronical, line 280:
```suggestion
    pub fn resize_shaped_words_cache(&self, capacity: usize) {
```

tronical:
Looks good to me. I don't think the `mut` is needed for the `self` receiver, but that's a small detail. @lemmih @adamnemecek what do you think?

tronical (approved):

PR #85, opened 2022-03-13
Allow breaking text in the middle of words when it is the first word in the paragraph

### LINKED ISSUE #34: Text breaking broken
If a text is too long to fit into a single line, splitting the word fails.
For example, let's assume, a single line can only contain three letters.

```
// the index of break_text will be 0. I would expect 3 (5 would be fine, too)
assert_eq!(canvas.break_text(width_of_three_letters, "Hello World", text_paint), Ok(0));

// just returns an empty Vec, I would expect vec!["Hel", "lo", "Wor", "ld"] (vec!["Hello", "World"] would be fine, too)
assert_eq!(canvas.break_text_vec(width_of_three_letters, "Hello World", text_paint), Vec::new());
```

So the main problem is, that text is never split in the middle of words (or an alternative: allow more letters than the maximum specified line width)

It's just a corner case, so it does not matter what it does exactly, as long as I get indices to all the text in some way.

rhelmot:
Closes #34 

Here's the picture worth a thousand words:

```
    let text = "This is longer chunk of text. aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\n\nWould have used lorem ipsum but she was busy jumping over the lazy dog with the fox and all the men who came to the aid of the party.🎉";
```

image

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -585,7 +585,7 @@ fn shape_run(
             let mut word_break_reached = false;
             let mut byte_index = run.start;
 
-            for word_txt in sub_text.split_word_bounds() {
+            for mut word_txt in sub_text.split_word_bounds() {
                 let id = ShapingId::new(paint, word_txt, max_width);
 
                 if !context.shaped_words_cache.contains(&id) {
@@ -631,9 +631,11 @@ fn shape_run(
                                 }
 
                                 if let Some(Ok(subword)) = context.shaped_words_cache.get(&id) {
-                                    // replace the outer variable so we can continue normally
+                                    // replace the outer variables so we can continue normally
                                     word = subword.clone();
-                                    byte_index += subword_txt.len();
+                                    word_txt = subword_txt;
+                                } else {
+                                    break;
                                 }
                             } else {
                                 // we are not breaking up words - discard this word
@@ -651,12 +653,13 @@ fn shape_run(
 
                     words.push(word);
                     first_word_in_paragraph = false;
-                    if word_break_reached {
-                        break;
-                    }
                 }
 
                 byte_index += word_txt.len();
+
+                if word_break_reached {
+                    break;
+                }
             }
 
             if levels[run.start].is_rtl() {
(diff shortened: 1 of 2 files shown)


rhelmot:
If you like, I can write up a proper solution to word wrapping using the [textwrap](https://mgeisler.github.io/textwrap/) crate and feature-gate it.

src/text.rs
```diff
@@ -596,7 +599,48 @@ fn shape_run(
...
+                                let mut subword_width = 0.0;
+                                let target_width = max_width - result.width;
+                                for glyph in word.glyphs.iter() {
+                                    bytes_included = glyph.byte_index;
+                                    let glyph_width = glyph.advance_x + paint.letter_spacing;
+
+                                    // nuance: we want to include the first glyph even if it breaks
+                                    // the bounds. this is to allow pathologically small bounds to
+                                    // at least complete rendering
+                                    if subword_width + glyph_width > target_width && bytes_included != 0 {
```
tronical, line 614: I wonder if this should be `>=`?

tronical (commented):
Looks good to me. Just one question inside :)

i think a feature gated text wrap makes sense.

src/text.rs
rhelmot: I don't think it matters, because we're comparing floats, but I've fixed it!

tronical: Yeah, it's more of a matter of consistency :). Thanks!

tronical (approved):

rhelmot:
I've pushed one more commit which is hopefully related enough to be included with this PR. This commit changes the line-breaking behavior when the final word in the line, that is, the word which crosses the maximum width, is whitespace. Previously, femtovg would make it so that the final whitespace would appear at the start of a new line, which I believe is not desirable behavior for any graphics application. My best guess as to what the correct behavior should be is that this word should be included in the byte count for the line but not in its metrics, i.e. its width.

Please let me know what you think.

tronical:
I agree the test is wrong. Please feel free to adjust it. A new line should not start with white space.

tronical (approved):

tronical:
Thanks:) and sorry it took so long :(

PR #90, opened 2022-05-02
Re-export image, imgref, and rgb

### LINKED ISSUE #88: Expanded image loading API?
I don't know if my use case for this is valid so I'll describe where I'm at and what I'd like and you can tell me if it seems like a good idea.

I would like to be able to load images in a background thread. Currently, to do this I have to depend on the exact same version of `image`, `rgb`, and `imgref` that femtovg does, and then rely on e.g. the `From<ImgRef<'a, RGBA8>>` impl to get myself a ImageSource, which I can ferry to the main thread and then load into my canvas lazily.

Instead, I would like to just be able to say `femtovg::ImageSource::load_from_rgba_buf(mem, pitch)` and `femtovg::ImageSource::load_from_mem(mem, femtovg::guess_format(mem))` - basically the same api as `image` and `imgref` provide. In fact, it might be best to cut out the middle man and just directly re-export `image` and `imgref` and `rgb`.

Let me know what you think!

rhelmot:
Closes #88 

If you would prefer to have the image module renamed so the image re-export can keep its common name, I will do that :)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1439,3 +1439,10 @@ impl<T: Renderer> Drop for Canvas<T> {
         self.images.clear(&mut self.renderer);
     }
 }
+
+// re-exports
+#[cfg(feature = "image-loading")]
+pub use ::image as img;
+
+pub use imgref;
+pub use rgb;


src/lib.rs
```diff
@@ -1439,3 +1439,10 @@ impl<T: Renderer> Drop for Canvas<T> {
         self.images.clear(&mut self.renderer);
     }
 }
+
+// re-exports
+#[cfg(feature = "image-loading")]
+pub use ::image as img;
```
tronical, line 1445: I’ve see other crates re-export the image crate as image. Is it possible to do this here, too?

rhelmot: It is not, at least not without some invasive albeit minor refactoring, since we already have a module named image. See my original post in this thread.

tronical: Ah yes. Sorry, I missed that :)

tronical (approved):

PR #95, opened 2022-05-28
Make new_from_glutin_context generic over all context types

whatisaphone:
Prepare for a flurry of PRs! 😄

This makes `new_from_glutin_context` more generic so it can be used with headless and surfaceless contexts

`ContextWrapper` derefs to `Context` so most existing callers of this function will keep working (including all the examples)


tronical (approved):
Excellent idea! Would be great if you could add an entry to the changelog. But I can also do that later :)

PR #96, opened 2022-05-28
Mark some Color functions as const

whatisaphone:
Also on my wishlist is creating color constants. I marked all the Color functions I could as `const`. Floating point math isn't supported yet, so many were skipped.


tronical (approved):
👍👍

PR #97, opened 2022-05-28
Change non-drawing text functions to shared self

whatisaphone:
Text layout is semantically non-mutating, so it would be nice if you didn't need a mut reference to the canvas.


tronical (approved):
Agreed

Issue #103, opened 2022-07-17
Text line splitting improvements


porky11:
Short list of problems:
- [ ] 1: symbols should belong to words and should not be split from them if possible
- [ ] 2: multiple symbols belong together and should not be split if possible
- [ ] ~3: spaces should be trimmed between split lines~

# Letters and symbols

In the current version of `femtovg` (0.3.5), text line splitting, for example using `canvas.break_text` doesn't work as you would expect.

For example if I have a sentence like this: "Sentence which ends with a dot." And the dot (".") does not fit in the line, while the word "dot" does, it will be split like this:

Line 1: "Sentence which ends with a dot"
Line 2: "."

I would expect it to be like that:

Line 1: "Sentence which ends with a"
Line 2: "dot."

The rule would be, if a letter is followed by a symbol, the symbol still belongs to that word.

# Multiple symbols

Same should be true for a sequence of multiple symbols (like ...).
This would be split into multiple lines as well, if it does not fit in.

For example a line like this "Hello ..."
Might be split into lines like this:

Line 1: "Hello ."
Line 2: ".."

In this case, I would expect this:

Line 1: "Hello"
Line 2: "..."

Combined this means if I have this sentence "He thinks: I would like to go, but..."

It currently could be like this:

Line 1: "He thinks: I would like to go, but.."
Line 2: "."

While it should be like this:

Line 1: "He thinks: I would like to go,"
Line 2: "but..."

# Spaces

Besides that there's a third problem:

If I have multiple words like this split using spaces: "Hello world!"

It could be split like this:

Line 1: "Hello " (see there's still a space after Hello"
Line 2: "World!"

It should rather be split like this:

Line 1: "Hello"
Line 2: "World!"

Else it might cause alignment issues when rendering.

(not sure about this last one. Manual trimming should also be possible, I guess)

geom3trik:
I think I mentioned this before in the Discord but I think one possibility to fix this issue is to use something like this crate <https://github.com/mgeisler/textwrap> with femtovg. The textwrap crate can use unicode breaks, which is a set of rules to determine which characters can be broken and which should stay together when breaking lines.

porky11:
@geom3trik 
I just implemented the rules I mentioned manually.
Wasn't too difficult. I'll push it after I did a bit of testing.

I don't think, textwrap is really suitable. It wraps based on character count and doesn't take in count that characters might not be monospace.
Maybe there is a way to take non monospace characters into account and I just didn't find one yet.
[...]

PR #104, opened 2022-07-17
Use custom segmentation method instead of unicode segmentation

porky11:
This will solve #103

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,6 @@ bitflags = "1.2.1"
 ttf-parser = "0.15.0"
 rustybuzz = "0.5.0"
 unicode-bidi = "0.3.4"
-unicode-segmentation = "1.6.0"
 generational-arena = "0.2.8"
 lru = { version = "0.7.1", default-features = false }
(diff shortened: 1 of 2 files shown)


tronical:
I’m not sure that this is the right path. Breaking between spaces is - in principle - the right approach for western languages. But it’s not for East Asian languages.

Unicode word boundaries are suboptimal for like breaking as there are line break opportunities that are not word boundaries.

I suspect that the primary motivation for word boundaries here is that it increases the chances of meeting reasonable (and cacheable) boundaries for shaping.

I think replacing the Unicode word boundaries algorithm with a hand made algorithm for line break boundaries that also comes with no test coverage is a regression.

I think it would be better to pick line breaks as per [unicode-linebreak](https://docs.rs/unicode-linebreak/0.1.2/unicode_linebreak/) opportunities, mandatory and optional ones.

That however may require changing or adapting the caching mechanism for the shaping.

I think a new API is required that allows shaping the text once, break it into multiple lines and the draw the glyphs, without shaping again.

The word caching avoids the re-shaping that would otherwise be required by the API as after the line breaking the shaper output is discarded.

PR #109, opened 2022-08-19
Remove ouroboros dependency

geom3trik:


tronical:
I think in principle it is desirable to eliminate the ourobos dependency, just to "clean house". The reason why it's there was to preserve the existing behavior of not parsing the truetype table headers for the bare true type fonts as well as shaping (rusty buzz) every time we need to add a glyph to the glyph cache. It feels really silly to do that every time. That said, I'm also not sure this is really a hot path. Ideally there'd be a way of batching the updates in a way that the temporary face could be created only once per batch of "new" glyphs.

That is my primary concern. Code wise I think this patch is fine.

geom3trik:
> The reason why it's there was to preserve the existing behavior of not parsing the truetype table headers for the bare true type fonts as well as shaping (rusty buzz) every time we need to add a glyph to the glyph cache.

Right I see. For some reason I had it in my mind that creating the face didn't actually do the parsing of the font tables but I've now checked and it does, so I see now the motivation for having the self referencing for when the font data is owned by the text context.

I think it would still be good to remove the ouroboros dependency. While researching this further and looking into the `owned_ttf_parser` crate (which iirc was being used before?), this crate also uses a self reference to data but with some kind of unsafe trick instead of ouroboros: <https://docs.rs/owned_ttf_parser/latest/src/owned_ttf_parser/owned.rs.html#109>.

I don't know how you feel about using some unsafe code here but I believe it to be sound and I suppose well tested by the fact that it's used in the `owned_ttf_parser ` crate.

tronical:
I think it's better to have unsafe code in depending crates like ourobous that are used by many others and likely to a wider review.

It might not be easy, but perhaps there *is* a way to re-organize the femtovg code to batch up the glyph updates.

PR #111, opened 2022-09-16
Shader example

ColRad:
Hey femtovg Team, 

I really like working with femtovg!
One thing that was missing for me was custom screen shaders. So I figured out a way to make it work using a framebuffer and `set_screen_target()`.

I thought an example would help others with similar needs.

### DIFF
--- a/examples/shader.rs
+++ b/examples/shader.rs
@@ -4,7 +4,6 @@ use glow::{Context, HasContext, NativeFramebuffer, NativeProgram, NativeTexture,
 use glutin::event::WindowEvent;
 use glutin::ContextBuilder;
 use glutin::{Api, GlRequest};
-use std::fs;
 use winit::event::Event;
 use winit::event_loop::{ControlFlow, EventLoop};
 use winit::window::WindowBuilder;
@@ -93,21 +92,21 @@ fn main() {
 fn create_shader_program(context: &glow::Context) -> NativeProgram {
     unsafe {
         let v_shader = context.create_shader(glow::VERTEX_SHADER).unwrap();
-        let vert_shader: String = fs::read_to_string("assets/screen.vert.glsl").unwrap().parse().unwrap();
-        context.shader_source(v_shader, vert_shader.as_str());
+        let vert_shader = include_str!("../assets/screen.vert.glsl");
+        context.shader_source(v_shader, vert_shader);
         context.compile_shader(v_shader);
         if context.get_shader_compile_status(v_shader) == false {
             let error_msg = context.get_shader_info_log(v_shader);
-            println!("ERROR::SHADER::VERTEX::COMPILATION_FAILED\n{:?}", error_msg);
+            panic!("ERROR::SHADER::VERTEX::COMPILATION_FAILED\n{:?}", error_msg);
         }
 
         let f_shader = context.create_shader(glow::FRAGMENT_SHADER).unwrap();
-        let frag_shader: String = fs::read_to_string("assets/screen.frag.glsl").unwrap().parse().unwrap();
-        context.shader_source(f_shader, frag_shader.as_str());
+        let frag_shader = include_str!("../assets/screen.frag.glsl");
+        context.shader_source(f_shader, frag_shader);
         context.compile_shader(f_shader);
         if context.get_shader_compile_status(f_shader) == false {
             let error_msg = context.get_shader_info_log(f_shader);
-            println!("ERROR::SHADER::FRAGMENT::COMPILATION_FAILED\n{:?}", error_msg);
+            panic!("ERROR::SHADER::FRAGMENT::COMPILATION_FAILED\n{:?}", error_msg);
         }
 
         let shader_program = context.create_program().unwrap();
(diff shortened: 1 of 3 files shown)


geom3trik:
Oh this is awesome! Thank you for this. I'm wondering if it might be a good idea to expose the neccessary glow types through femtovg so the user doesn't have to drop down to glow directly? @tronical what are your thoughts on this?

tronical:
> Oh this is awesome! Thank you for this. I'm wondering if it might be a good idea to expose the neccessary glow types through femtovg so the user doesn't have to drop down to glow directly? @tronical what are your thoughts on this?

I think it’s better of glow remains an implementation detail. This way glow can be changed to new versions at convenience without a major version bump, for example. New glow versions tend to change API in ways that typically affect I think very very few applications, but it still needs to be a new major version for femtovg.

I think it’s perfectly fine for application code to freely choose their glow version (I was recently picking a git master version for an example) - everyone shares the same GL context underneath.

In this case the change to example would seem insignificant to me: use a re-exported glow module from femtovg or just depend on the glow crate.

tronical:
I don’t know do the API for it still works, but one modification I would suggest for this example (good idea btw!) is this:

The OpenGl renderer type that is created has a set_screen_target() function that can take an FBO. I suggest to use that, render with femtovg as usual, flush() and then render the FBO.

This way the boundary between femtovg and the rest of the GL state is a bit clearer. The trick of initializing femtovg and then under its feet bind the current framebuffer to an FBO works but it’s a little hackish IMO compared to using the API.

What do you think?

ColRad:
Hey, thanks for the responses!

Honestly I'm punching a bit above my weight here. I'm usually a webdev. 

@tronical I'm unsure what you are pointing at 
> I suggest to use that, render with femtovg as usual, flush() and then render the FBO.

`OpenGL` only has 2 public initialization methods: `new_from_function()` & `new_from_glutin_context()`
It seems like I have to create `femtovg` with some kind of render target. And then replace it via `set_screen_target()` afterwards.

As for the rendering itself. Am I not doing it the way you describe right now? 
`L70` Clears the FBO
`L80` Render `femtovg` to FBO
`L83` Render FBO to Window

image

Sorry for all the questions. I'm quite rusty in OpenGL. Last time was modding Q3A in High school ;)

examples/shader.rs
```diff
@@ -0,0 +1,194 @@
...
+            },
+            Event::RedrawRequested(_) => {
+                // draw red rectangle on white background
+
+                let dpi_factor = windowed_context.window().scale_factor();
+                let size = windowed_context.window().inner_size();
+                canvas.set_size(size.width as u32, size.height as u32, dpi_factor as f32);
+                canvas.clear_rect(0, 0, size.width as u32, size.height as u32, Color::rgbf(1., 1., 1.));
+
+                prepare_framebuffer_for_render(&context, framebuffer);
```
tronical, line 70: Suggestion to swap things around, so that prepare_framebuffer_for_render() prepares the scene for use with the femtovg canvas, which then does its thing, ends with `flush()` and then `render_framebuffer_to_screen` takes over.

```suggestion
                prepare_framebuffer_for_render(&context, framebuffer);

                // draw red rectangle on white background

                let dpi_factor = windowed_context.window().scale_factor();
                let size = windowed_context.window().inner_size();
                canvas.set_size(size.width as u32, size.height as u32, dpi_factor as f32);
                canvas.clear_rect(0, 0, size.width as u32, size.height as u32, Color::rgbf(1., 1., 1.));
```

examples/shader.rs
```diff
@@ -0,0 +1,194 @@
...
+            _ => (),
+        }
+    });
+}
+
+fn create_shader_program(context: &glow::Context) -> NativeProgram {
+    unsafe {
+        let v_shader = context.create_shader(glow::VERTEX_SHADER).unwrap();
+        let vert_shader: String = fs::read_to_string("assets/screen.vert.glsl").unwrap().parse().unwrap();
+        context.shader_source(v_shader, vert_shader.as_str());
```
tronical, line 97: Can you try and see if this trick works? I think that it might be more idiomatic to include the shaders at compile time. The same could be done for the fragment shader as well then.

```suggestion
        let vert_shader = include_str!("assets/screen.vert.glsl");
        context.shader_source(v_shader, vert_shader);
```

examples/shader.rs
```diff
@@ -0,0 +1,194 @@
...
+        context.compile_shader(v_shader);
+        if context.get_shader_compile_status(v_shader) == false {
+            let error_msg = context.get_shader_info_log(v_shader);
+            println!("ERROR::SHADER::VERTEX::COMPILATION_FAILED\n{:?}", error_msg);
```
tronical, line 101: Perhaps it would make sense here to use `panic!`, as this is overall quite fatal to the example :)

examples/shader.rs
ColRad: Yeah, that works! Nice!

examples/shader.rs
ColRad: Makes sense ;)

tronical (approved):

PR #114, opened 2022-10-03
Refactor of examples

porky11:
All examples besides shaders run on web now.

Window creation is moved to a new module.
This is also done to simplify porting the examples to different backends.


tronical:
Hi! If possible, please rebase your branch to the latest master. Then it’ll be easier to see the actual changes :)

porky11:
@tronical Rebase complete.

porky11:
Here the web examples can be tested now: https://45.76.84.21/femtovg_demo/

porky11:
CI failed. I have to find some way to ignore examples.

porky11:
@geom3trik 
Thanks, pushed.

Yeah, the tests on my website were just a quick way so you can have a look, and I'll probably take it down again one day. Something on an official website would be nice, too.

porky11:
It's more than just getting every example work on web. Most importantly it's a refactor and cleanup of the examples.

porky11:
@tronical Rebased to master, what do you think?

examples/gradients.rs
```diff
@@ -1,33 +1,35 @@
...
-        windowed_context.window().scale_factor() as f32,
-    );
+#[cfg(target_arch = "wasm32")]
+use winit::window::Window;
+
+fn run(
+    mut canvas: Canvas<OpenGl>,
+    el: EventLoop<()>,
+    #[cfg(not(target_arch = "wasm32"))] windowed_context: glutin::WindowedContext<PossiblyCurrent>,
+    #[cfg(target_arch = "wasm32")] window: Window,
```
tronical, line 29: The "#ifdeffery" here is quite painful, but I also see how this isn't easily addressed. The only way that I could imagine is to wrap event loop, context and (optional) Window in another type and let the example provide a callback that is called from `run()` and which (unconditionally) takes the `&WindowedContext`, `&Window` and `&Event`.

tronical (approved):
I think it would be really nice to find a cleaner solution for the cfg-stuff in the examples, but at the same time this patch is IMO moving into the right direction of making the examples web-accessible :).

Let me know if you'd like to work further on encapsulation of the #cfg stuff or if you prefer landing this.

porky11:
Most of them could be removed and will lead to unused import or unused arguments warnings, which could instead be replaced by `#[allow(unused...)]` stuff, or just be ignored. Probably a matter of preference.

I'm not sure if I like a refactor to make it even more complicated in the examples. People want to use them as a base, so having multiple files is already a change I'm not totally happy with. I think in this case it's acceptable, since window creation is required in every example.

For now I don't have interest to work on this anymore, so I'd prefer landing it.

I'd also like to see a new release with updated dependencies. There's just one thing which would be useful, which might be a breaking change: A new feature for OpenGL, which is only enabled if the opengl renderer should be used.
But if this is in the default features, it probably isn't a breaking change, so it might also be done in a minor release, I guess.

tronical:
> For now I don't have interest to work on this anymore, so I'd prefer landing it.

👍 

> I'd also like to see a new release with updated dependencies. There's just one thing which would be useful, which might be a breaking change: A new feature for OpenGL, which is only enabled if the opengl renderer should be used. But if this is in the default features, it probably isn't a breaking change, so it might also be done in a minor release, I guess.

Making an existing feature optional is a breaking change though. There's a solution though: https://slint-ui.com/blog/rust-adding-default-cargo-feature.html

I'm in favor of using this approach and with the new major release we have the opportunity of introducing it.

porky11:
>There's a solution though

I already mentioned default features.

PR #117, opened 2022-10-10
Huge paint refactor

porky11:
Many commits are pretty specific, so here a list of general changes:

* supply paint as reference in exposed functions
* only store things in paint, which can be set manually, remove temporary data (`transform` and `glyph_context` have been removed)
* only supply parameters to functions which are used instead of supplying the whole paint struct (for example `GlyphContext` always set to default except when calling `render_triangles`, the only case where it even could be a different value)
* grouped common stuff in paint struct into new sub structs (stroke-only parameters in `StrokeSettings`, text only parameters in `TextSettings`)
* avoid copying full paint struct (only the newly created sub structs are copied sometimes)


porky11:
I still like the idea of a API, where you supply individual structs.

The four main drawing function type signatures would look somewhat like this:

```
pub struct FillPaint {
    pub flavor: PaintFlavor,
    pub shape_anti_alias: bool,
    pub fill_rule: FillRule,
}
pub fn fill_path(&mut self, path: &mut Path, paint: &FillPaint);

pub struct StrokePaint {
    pub flavor: PaintFlavor,
    pub shape_anti_alias: bool,
    pub stroke: StrokeSettings,
}
pub fn stroke_path(&mut self, path: &mut Path, paint: &StrokePaint);

pub struct TextFillPaint {
    pub flavor: PaintFlavor,
    pub shape_anti_alias: bool,
    pub fill_rule: FillRule,
    pub text: TextSettings,
}
pub fn fill_text<S: AsRef<str>>(&mut self, x: f32, y: f32, text: S, paint: &TextFillPaint) -> Result<TextMetrics, ErrorKind>;

pub struct TextStrokePaint {
    pub flavor: PaintFlavor,
    pub shape_anti_alias: bool,
    pub stroke: StrokeSettings,
    pub text: TextSettings,
}
pub fn stroke_text<S: AsRef<str>>(&mut self, x: f32, y: f32, text: S, paint: &TextStrokePaint) -> Result<TextMetrics, ErrorKind>;
```

(see the current definitions of `StrokeSettings` and `TextSettings` in this pull request)

There would still be the generic paint struct like it's defined here, which can be supplied to all the functions.
(the API would have to change a bit, but it should be possible)

Or maybe just expose the lower level API directly, which just takes in different numbers of parameters, and let the user define abstractions if they need them.

The main reasoning: If the API is lower level, I can decide if I need abstractions, and how they look like.
For example if I create some vector graphics format, and just use the current paint to write data to my format, my format will contain a lot of junk. I always have the option to create my own structs or my own serializers, but just writing everything inside the exposed struct might also have benefits.

porky11:
I'm pretty happy with it now. I also did some general cleanup and fixed serde serialization in the end.

src/text.rs
```diff
@@ -373,15 +386,15 @@ impl TextContextImpl {
         self.fonts.get_mut(id.0)
     }
 
-    pub fn find_font<F, T>(&mut self, paint: &Paint, mut callback: F) -> Result<T, ErrorKind>
+    pub fn find_font<F, T>(&mut self, font_ids: [Option<FontId>; 8], mut callback: F) -> Result<T, ErrorKind>
```
tronical, line 389: I think just passing a reference to a slice would be sufficient here, as well as in many other places.

src/text.rs
```diff
@@ -756,13 +792,13 @@ fn layout(
     y: f32,
     context: &mut TextContextImpl,
     res: &mut TextMetrics,
-    paint: &Paint,
+    text: &TextSettings,
```
tronical, line 795: text is I think a typical variable name of a string (like i is for loop iteration). I suggest to rename to text_settings.

tronical (commented):
Awesome, I like the direction!

src/text.rs
porky11: Mainly because I copied it from the type from struct.

src/text.rs
porky11: Yeah, I also had problems because of that in some functions and already named it text_settings there.

src/text.rs
porky11: I store font ids as `Vec` now, not sure if this is a good idea.
I also used `Rc` to avoid too much copying.
At least when building the cache, there's still a lot of cloning required.

Not sure if it's better than the suggested change?

porky11:
I reverted the changes to use `Vec` instead of `[Option<FontId>; u8]`, I just added references in some places.

tronical (approved):
Looks good to me. A ChangeLog entry would be nice, too. But we can do that before the actual release, too.

porky11:
@tronical Changelog done. I hope the phrasing is good enough.

tronical:
Looks good - thanks! Shall I merge?

porky11:
@tronical Yeah

PR #123, opened 2022-11-03
Fixes a sign error introduced in #9da49b71dcbe681e3358cd6410f30ce4eb7a5732

geom3trik:


tronical (approved):

tronical (changes requested):
Looks good to me, but please change the commit message to match the PR instead of just “change cache.rs” as that’s what’ll be in the history :)

tronical (approved):
Let’s see if this can be solved via squash…

PR #126, opened 2022-11-07
If let refactor

porky11:
I replaced a few if-let and match statements by functional style and the new let-else syntax.
Increases the MSRV to the current version (1.65.0), which should be considered before merging.

### DIFF
--- a/examples/demo.rs
+++ b/examples/demo.rs
@@ -958,11 +958,10 @@ fn draw_button<T: Renderer>(
     paint.set_text_align(Align::Left);
     paint.set_text_baseline(Baseline::Middle);
 
-    let tw = if let Ok(layout) = canvas.measure_text(0.0, 0.0, text, &paint) {
-        layout.width()
-    } else {
-        0.0
-    };
+    let tw = canvas
+        .measure_text(0.0, 0.0, text, &paint)
+        .map(|layout| layout.width())
+        .unwrap_or(0.0);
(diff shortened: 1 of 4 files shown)


geom3trik:
I like the refactor in general, but I think we should wait a few more Rust versions because this will push all dependant crates, like my own, up to MSRV 1.65 and users tend to be slow to update their Rust compiler.

examples/demo.rs
```diff
@@ -958,11 +958,10 @@ fn draw_button<T: Renderer>(
...
 
-    let tw = if let Ok(layout) = canvas.measure_text(0.0, 0.0, text, &paint) {
-        layout.width()
-    } else {
-        0.0
-    };
+    let tw = canvas
+        .measure_text(0.0, 0.0, text, &paint)
+        .map(|layout| layout.width())
+        .unwrap_or(0.0);
```
tronical, line 964: I think this can be simplified:

```suggestion
        .map_or(0.0, |layout| layout.width())
```

tronical (commented):
I think requiring a newer Rust in the example code makes sense, but I share the hesitancy towards bumping the MSRV for crate users.

examples/demo.rs
porky11: Thanks, I didn't know about that yet. Clippy already told me about such simplifications multiple times.

Issue #134, opened 2022-12-24
Using femtovg with swash


rhelmot:
[swash](https://github.com/dfrg/swash) is a new library for text shaping/scaling/rasterization. In particular, it is the output format of the also-new [COSMIC text](https://pop-os.github.io/cosmic-text/cosmic_text/index.html) library, which seems to be an extremely strong base for text layout and editing. From what I can tell, swash can provide vector rendering instructions in the form of [zeno](https://docs.rs/zeno/0.2.2/zeno/index.html) objects, but it can also just do the rasterization itself.

- do we want to replace femtovg's text layout with COSMIC text?
- do we want to replace femtovg's shaping algorithm with swash?
- do we want to provide a public interface for rendering zeno vectors? swash bitmaps?

I will be working on the translation layer from swash image bitmaps to bitmaps that femtovg can consume. This will live in vizia for the time being, but if you want it I can PR it :)

tronical:
I’m in favor of cosmic, in combination with rustybuzz as shaper (but perhaps that can remain configurable for users of the crate, like with cosmic). Also IMO yes to a public interface for “supplying glyph representations”.

rhelmot:
Skeleton, untested proof of concept, soliciting feedback on what parts of this are worth PRing in what form:

- [femtovg diffs](https://github.com/rhelmot/femtovg/commit/e1f3f05cc050c0ed69ca702b368bb5d4eb2cabff)
- [client code utilizing those diffs](https://github.com/vizia/vizia/pull/275/commits/dc9a771b1036502ef874cba1d95a60c0a1b54af4#diff-60e92b6a7ba2848b2c7159ade25257558d17a68ac73b1e691a6ffa44b3d466db)

I picked the way to modify femtovg that required the absolute minimum effort, so obvi not set in stone, but everything typechecks!

tronical:
Since 0.20.2 swash rendering is supported via the Cargo feature with the same name :)

PR #135, opened 2023-01-02
Export a bunch of interfaces to allow external font rendering

rhelmot:
As foretold in #134. This is essentially the smallest possible refactor I could make in order to externally manage font atlases and push them into femtovg.

If for any reason this is an unsavory change, let me know and I will do my best to make it more palatable.

[Sample usage](https://github.com/vizia/vizia/blob/8359c4dfe0234ba53ed245a42612e9e0967d46c4/crates/vizia_core/src/text/cosmic.rs#L124), now merged into Vizia's main branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,6 +62,9 @@ image = { version = "0.24.0", default-features = false, features = [
     "jpeg",
     "png",
 ] }
+cosmic-text = { git = "https://github.com/pop-os/cosmic-text", rev = "e00109d77f06d5a2e3057865eda3f530bc40a046" }
+swash = "^0.1"  # keep this in sync with cosmic-text
+lazy_static = "1.4.0"
 
 [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
 glutin = "0.29.0"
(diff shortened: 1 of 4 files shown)


tronical:
Interesting approach! Do I understand correctly that this basically also delegates the glyph atlas texture handling to the caller?

tronical (approved):
Thanks for the explanation. I quite like this :). It's really minimal and also only exposes the texture format. I think what would really help is documentation (about the expected format of the texture, the coordinate system etc).

Given that nobody else commented I suppose there are no other objections:). Want me to merge as-is?

rhelmot:
I added an example which should hopefully illustrate the bits you were interested in.

The example panics if you resize the window too small, but this is a bug in cosmic-text.

rhelmot:
Oh, if you were directing that question at me: yes, you can merge this as-is.

tronical:
Looks like there’s a build issue in cosmic-text. Is that MSRV related?

rhelmot:
Testing indicates cosmic-text has an undeclared MSRV of 1.63, but for the life of me I don't understand what the problem is and what from the 1.63 changelogs is fixing it. So we have a few options:

- Bump femtovg MSRV to 1.63
- Help me figure out how to fix cosmic-text for 1.61
- Move the example to a pending pull request which will be merged when one of the above is satisfied

tronical:
I bumped the MSRV to 1.63 in femtovg master branch. Can you rebase your PR?

PR #137, opened 2023-01-22
Move scale for glyph commands to function argument

geom3trik:
Fixes a bug where text rendered with `draw_glyph_cmds`, but not with `draw_text`, would be incorrect when transformed. This change allows passing a custom scale to the function, which should be `1.0` when used externally.

### DIFF
--- a/examples/external_text.rs
+++ b/examples/external_text.rs
@@ -258,7 +258,7 @@ fn run(
                 let cmds = cache
                     .fill_to_cmds(&FONT_SYSTEM, &mut canvas, &buffer, (0.0, 0.0))
                     .unwrap();
-                canvas.draw_glyph_cmds(cmds, &Paint::color(Color::black()));
+                canvas.draw_glyph_cmds(cmds, &Paint::color(Color::black()), 1.0);
 
                 canvas.flush();
                 #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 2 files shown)


tronical (approved):

rhelmot (approved):

rhelmot:
I definitely copy-pasted the draw_glyph_commands code without understanding what it did, but this seems sound. @tronical if you have any advice on why text has a separate scaling parameter I'd love to hear it.

tronical:
> I definitely copy-pasted the draw_glyph_commands code without understanding what it did, but this seems sound. @tronical if you have any advice on why text has a separate scaling parameter I'd love to hear it.

I'm not entirely sure I understand the question. Do you mean to ask why this function exists?

```rust
    fn font_scale(&self) -> f32 {
        let avg_scale = self.state().transform.average_scale();

        geometry::quantize(avg_scale, 0.1).min(7.0)
    }
```

rhelmot:
Yes. Why is it important to quantize the current scale level and scale text with respect to that?

peter-todorov:
It's because glyphs are cached in an atlas. And since the font size is affected by the current transform scale, one can generate a huge number of glyph sizes, for example if there's a scroll to zoom implemented in the client application. Quantizing the size makes the possible number of glyph sizes smaller and the atlas cache more useful.

PR #150, opened 2023-02-17
Expose the structures needed to implement your own renderer outside the crate

sumibi-yakitori:
I would like to implement my own Renderer, and to do so I would like these structures to be exposed outside the crate.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,8 +37,7 @@ pub use text::{
 use text::{GlyphAtlas, TextContextImpl};
 
 mod image;
-use crate::image::ImageStore;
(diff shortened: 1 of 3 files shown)


porky11:
I never thought of it, but being able to separate the specific renderers into their own crates might be useful.
This way, the current opengl renderer could also get its own crate.

sumibi-yakitori:
@porky11 
It seems that this PR is incomplete and there are still many structures that need to be exposed in order to make it complete.
Furthermore, I have a feeling that those structures are likely to change in the future, so I have given up on the idea of creating my own renderer.

PR #162, opened 2023-04-25
Change set_size to accept anti-alias-scale.

### LINKED ISSUE #160: Decouple DPI and text anti-alias in Canvas
https://github.com/femtovg/femtovg/blob/7ac3df83e92966243c6c0c648cdd989682e14072/src/lib.rs#L368-L371

This is an issue to solve the blurry text on Windows of slint. Slint handles HiDPI itself, but the text rendering needs a larger value dpi for anti-alias on Windows. It would be better if we can pass two values, one for the scale factor in `measure_font`, and other for anti-alias.

Related: https://github.com/slint-ui/slint/issues/2615 https://github.com/slint-ui/slint/pull/2618

Berrysoft:
Fixes #160 

This will cause a breaking change. Should we choose another type of API?

There are some other solutions:
1. Add another `set_size_with_anti_alias` method.
2. Add another `set_anti_alias` method.
3. Make the third parameter of `set_size` accept both `f32` and `(f32, f32)`. This could be done by trait magic.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -363,11 +363,17 @@ where
 
     /// Sets the size of the default framebuffer (screen size)
     pub fn set_size(&mut self, width: u32, height: u32, dpi: f32) {
+        self.set_size_with_anti_alias(width, height, dpi, dpi.ceil());
+    }
+
+    /// Sets the size of the default framebuffer (screen size) and anti-alias scale.
+    /// The `anti_alias_scale` could be any value, but usually larger than or equal to `dpi`.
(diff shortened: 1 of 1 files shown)


Berrysoft:
I edited the patch to avoid breaking changes.

I don't know how to describe the parameter either. From the apperance, I would say that the larger the clearer, and it should be larger than or equal to the DPI scale.

If we change the parameter to an `enum Quality`, what should the three choices actually mean? I would suggest that
* `Low` means `anti_alias_scale == 1.0`;
* `Medium` means `anti_alias_scale == 2.0`;
* `High` means `anti_alias_scale == 3.0`.
Is that suggestion OK?

PR #184, opened 2023-11-12
Add method for retrieving the native texture of an image

geom3trik:
I'm using this for sharing textures between windows in a multi-window setup. The texture can be created and managed by the main window canvas and then this method (`get_native_texture`)  can be used to get the native texture, which can then be used with `create_image_from_native_texture` for the canvases of the  subwindows. This avoids needing to create the native texture manually, which avoids needing to import glow.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -551,9 +551,11 @@ where
         Ok(id)
     }
 
-    pub fn get_native_texture(&self, id: ImageId) -> Option<T::NativeTexture> {
+    /// Returns the native texture of an image given its ID.
+    pub fn get_native_texture(&self, id: ImageId) -> Result<T::NativeTexture, ErrorKind> {
         self.get_image(id)
-            .and_then(|image| self.renderer.get_native_texture(image).ok())
+            .ok_or(ErrorKind::ImageIdNotFound)
+            .and_then(|image| self.renderer.get_native_texture(image))
     }
 
     pub fn get_image(&self, id: ImageId) -> Option<&T::Image> {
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -551,6 +551,11 @@ where
         Ok(id)
     }
 
+    pub fn get_native_texture(&self, id: ImageId) -> Option<T::NativeTexture> {
```
tronical, line 554: I think it would be nice to have some documentation. (I know get_image doesn’t, but trying to start a counter-trend ;)

As a caller I’d like to know what it means when this fails or returns None.

Perhaps this function should return a Result instead of Option?

tronical (commented):
Looks good to me. A few API nitpicks - curious what you think.

src/renderer.rs
```diff
@@ -104,6 +104,7 @@ pub trait Renderer {
     ) -> Result<Self::Image, ErrorKind>;
     fn update_image(&mut self, image: &mut Self::Image, data: ImageSource, x: usize, y: usize)
         -> Result<(), ErrorKind>;
+    fn get_native_texture(&self, image: &Self::Image) -> Result<Self::NativeTexture, ErrorKind>;
```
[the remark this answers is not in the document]
geom3trik: Ah yes, good idea!

src/lib.rs
geom3trik: No problem, I'll add some docs. And good idea on making it return a result.

tronical (approved):

tronical:
@geom3trik Added you to the committers. Feel free to rebase and merge.

PR #191, opened 2023-12-29
Improve performance of large texts

Drewol:
Reduces excessive calls to `face_ref` when drawing text directly. 

I used the std HashMap, does femtovg have a different preferred Map that should be used instead?

`face_ref` has allocations and does some sorting so it can be slow, to me it seems like it should only really be necessary to call this once per font in a more "global" context as even after this patch it is still called every frame. Could be something for the future but I have not looked at how.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -1,6 +1,7 @@
 use std::{
     borrow::Borrow,
     cell::RefCell,
+    collections::HashMap,
     ffi::OsStr,
     fs,
     hash::{Hash, Hasher},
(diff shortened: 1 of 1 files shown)


tronical:
Cool idea! Could you rebase the PR?:)

Drewol:
Sorry, I don't quite understand, where should I rebase it to?

tronical (approved):
Ah, my bad, all good, there's no need to rebase. Thanks for the patch! Makes a lot of sense.

PR #195, opened 2024-01-18
Fix to use log crates for logging

sumibi-yakitori:
Thanks for the great crate.

I am using this crate on macOS and GTK4 and femtovg outputs a lot of errors to stderr when I resize windows and such.
For me this behavior is causing me to miss other important logs.

Therefore, this PR will be changed to use the `log` crate, which allows you to adjust the log output.

https://github.com/femtovg/femtovg/assets/1794232/0d26091b-b1b0-4a45-bc96-1eeaa3ec00c7


tronical (approved):
Thanks:)

PR #206, opened 2024-04-08
Correctly calculate line width for all glyphs when stroking text in `text::render_direct()`

### LINKED ISSUE #205: Text stroke is incorrectly sized after rendering first glyph with direct rendering
I'm running into an odd issue with text strokes where occasionally the stroke width will be wrong after the first glyph.
Image of text with buggy outline
Same as above with no fill

I did some debugging and determined that the issue crops up whenever the text is rendered with `text::render_direct()`. The issue is this snippet:
https://github.com/femtovg/femtovg/blob/7bdd8e286de4e0758dde25f3f7cbc9a7cf200fc6/src/text.rs#L1268-L1273

I'm guessing the intent was to avoid repeatedly dividing the line width by the scale factor. However, because `line_width` is rebound every loop iteration, the effect is that the division only occurs for the first glyph, then does not occur for the rest.

I'll submit a PR shortly with a fix.

nununoisy:
Fixes #205.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -1242,7 +1242,6 @@ pub(crate) fn render_direct<T: Renderer>(
     let text_context = canvas.text_context.clone();
     let text_context = text_context.borrow_mut();
 
-    let mut scaled = false;
     let mut face_cache: HashMap<FontId, rustybuzz::Face> = HashMap::default();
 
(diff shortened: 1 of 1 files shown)


tronical (approved):
Good catch!

PR #215, opened 2024-09-25
Completed Documentation with help of AI

porky11:
Documentation is complete now. So no breaking changes.

I still thought, I won't merge it already in case anybody wants to have a look.
Mostly because I also changed existing documentation, especially for consistency (it's not perfect yet).

But I'll probably just merge it myself tomorrow.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -16,40 +16,53 @@ use crate::{ErrorKind, Renderer};
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct ImageId(DefaultKey);
 
-/// Image format: `Rgb8`, `Rgba8`, `Gray8`.
+/// Specifies the format of an image's pixels.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub enum PixelFormat {
+    /// 24-bit RGB image format (8 bits per channel)
     Rgb8,
+    /// 32-bit RGBA image format (8 bits per channel, including alpha)
     Rgba8,
+    /// 8-bit grayscale image format
     Gray8,
 }
 
 bitflags! {
-    /// Image flags (eg. repeat, flip, mipmaps, etc.)
+    /// Represents a set of flags that modify the behavior of an image.
     #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
     pub struct ImageFlags: u32 {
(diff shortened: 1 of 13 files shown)


src/image.rs
```diff
@@ -16,40 +16,53 @@ use crate::{ErrorKind, Renderer};
...
     Rgb8,
+    /// 32-bit RGBA image format (8 bits per channel, including alpha)
     Rgba8,
+    /// 8-bit grayscale image format
     Gray8,
 }
 
 bitflags! {
-    /// Image flags (eg. repeat, flip, mipmaps, etc.)
+    /// Represents a set of flags that modify the behavior of an image.
```
tronical, line 31:
```suggestion
    /// Represents a set of flags that modify the behavior of an image when rendering.
```

src/image.rs
```diff
@@ -16,40 +16,53 @@ use crate::{ErrorKind, Renderer};
...
     pub struct ImageFlags: u32 {
-        const GENERATE_MIPMAPS = 1;     // Generate mipmaps during creation of the image.
-        const REPEAT_X = 1 << 1;        // Repeat image in X direction.
-        const REPEAT_Y = 1 << 2;        // Repeat image in Y direction.
-        const FLIP_Y = 1 << 3;          // Flips (inverses) image in Y direction when rendered.
-        const PREMULTIPLIED = 1 << 4;   // Image data has premultiplied alpha.
-        const NEAREST = 1 << 5;         // Image interpolation is Nearest instead Linear
+        /// Generates mipmaps during the creation of the image.
+        const GENERATE_MIPMAPS = 1;
+        /// Repeats the image in the X direction when rendered.
```
tronical, line 36:
```suggestion
        /// Repeats the image in the X direction when rendered into an area wider than the image.
```

(same for the other)

src/image.rs
```diff
@@ -209,9 +224,7 @@ impl<T> ImageStore<T> {
         Ok(ImageId(self.0.insert((info, image))))
     }
 
-    ///
-    /// Reallocates the image without changing the id.
-    ///
+    // Reallocates the image without changing the id.
```
tronical, line 227: Shouldn’t this be triple slash?

tronical (approved):

src/image.rs
porky11: The type `ImageStore` isn't exported, so it doesn't need to be a doccomment. Could also be removed completely.

PR #221, opened 2024-11-10
Fix glyph overlap clipping

### LINKED ISSUE #183: Inter font is rendered incorrectly
Hi, I noticed that some characters from the Inter font are rendered incorrectly.

https://fonts.google.com/specimen/Inter

image

Some things to note:

- Characters `f`, `e`, `t` and `4` are rendered incorrectly (and more, I would imagine)
- It's a very popular font, so I do not think the font itself is at fault.
- Happens at any font size.
- The gaps get larger when a bolder variant of the font is chosen.
- Turning on/off anti-aliasing has no effect.
- I'm on Windows 11.
- Happens on macOS too it seems.

## Reproduction

1. Drop [`Inter-Medium.ttf`](https://fonts.google.com/specimen/Inter) into `examples/assets/`
2. In `examples/text.rs`, append `0123456789` to the `The quick brown fox jumps over the lazy dog` string
3. In `examples/text.rs`, change the `sans` font from `Roboto-Medium.ttf` to `Inter-Medium.ttf`.
4. `cargo run --example text`

richardhozak:
Fixes #183

Text was being rendered with wrong winding rule which caused overlapping glyph paths to create holes inside itself.

The issue is easily reproduced by using Inter font as described in #183 by swapping it for Roboto in demo example but applies to other fonts which are designed that way. The issue is most noticeable when zooming on text in demo (scaling the whole view).

Before:
Screenshot from 2024-11-10 20-53-45

After:
Screenshot from 2024-11-10 21-42-02

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -1113,7 +1113,7 @@ impl GlyphAtlas {
                             },
                         );
                     } else {
-                        canvas.fill_path_internal(path, &PaintFlavor::Color(mask_color), false, FillRule::EvenOdd);
+                        canvas.fill_path_internal(path, &PaintFlavor::Color(mask_color), false, FillRule::NonZero);
                     }
 
                     canvas.restore();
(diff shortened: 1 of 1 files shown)


tronical (approved):
This makes sense to me.

From https://developer.apple.com/fonts/TrueType-Reference-Manual/RM01/Chap1.html#direction :
> In making this determination, TrueType uses the non-zero winding number rule.

porky11 (approved):

PR #223, opened 2024-11-16
Update cosmic text to 0.12.1

richardhozak:
Updates cosmic text to latest version and updates external_text example accordingly.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -60,7 +60,7 @@ image = { version = "0.25.0", default-features = false, features = [
     "jpeg",
     "png",
 ] }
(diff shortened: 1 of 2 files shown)


richardhozak:
Note that I've noticed regression on cosmic text side with what I guess is invalid font fallback, now some emojis get rendered differently in their non-colored version. I've tested this with iced gui framework which also uses cosmic text and it has the same issue.

image

richardhozak:
I've also created a fix for the emoji rendering problem in cosmic-text here https://github.com/pop-os/cosmic-text/pull/328, but I do not think it should block this PR, we can update cosmic-text again later.

tronical (approved):

tronical:
Thx:)

PR #224, opened 2024-11-17
Simplify external text example and improve it's text rendering

richardhozak:
The main simplifications are:

- Using SwashCache to do the rasterization instead of doing it manually with swash
  - This also indirectly allows creating glyphs with fake italics if someone needs it
- Using nearest texture filtering for glyph textures which eliminates glyph padding and margin constants
  - This makes the glyph rendering also much better, see screenshots below

Without this PR:
Screenshot from 2024-11-17 10-57-29

With this PR:
Screenshot from 2024-11-17 10-55-42

Here's algo a gif switching between before and after, it might be easier to spot the differences (for me its most noticeable on the last emoji):
comparison

### DIFF
--- a/examples/external_text.rs
+++ b/examples/external_text.rs
@@ -1,6 +1,6 @@
 mod helpers;
 
-use cosmic_text::{Attrs, Buffer, CacheKey, FontSystem, Metrics, Shaping, SubpixelBin};
+use cosmic_text::{Attrs, Buffer, CacheKey, FontSystem, Metrics, Shaping, SubpixelBin, SwashCache};
 use femtovg::{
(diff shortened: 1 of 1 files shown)


tronical (approved):
Excellent. Makes sense to me. I'm dreaming of cosmic text as a built-in femtovg feature.

richardhozak:
Yeah, but the text rendering in femtovg is also cool as it is done with paths instead of rasterized glyphs from what I've gathered. We could probably use swash directly to provide path for femtovg to draw.

tronical:
The quality of the rasterization is a frequent issue we're seeing and I think swash is providing better quality.

Nanovg uses freetype iirc, which is the correct choice for C.

Cosmic text offers better don't fallback handling out of the box, and better line breaking.

Overall, I see enough value in using cosmic by default in femtovg.

Issue #229, opened 2025-01-07
Errors running examples in browser with wgpu feature


joshua-burbidge:
The Readme describes how to run the examples in a browser with WASM - is the same process supposed to work when the `wgpu` feature is enabled? When I attempt to run them I'm getting errors. Everything works as expected when using the OpenGL renderer with WASM, and when running with `cargo run --example demo --features wgpu`.

Commands from readme with `--features wgpu` added:
```
cargo install wasm-bindgen-cli
cargo build --target=wasm32-unknown-unknown --example demo --features wgpu
wasm-bindgen ./target/wasm32-unknown-unknown/debug/examples/demo.wasm --out-dir examples/generated --target web

cd examples/
python3 -m http.server
```

Result:
![image](https://github.com/user-attachments/assets/65fdc0ac-0340-4f89-b867-ad97cc0cf7ee)

```
Pipeline overridable constant ""shader_type"" not found in [ShaderModule "wgpu/shader.wgsl"].
 - While validating vertex stage ([ShaderModule "wgpu/shader.wgsl"], entryPoint: "vs_main").
 - While validating vertex state.
 - While calling [Device].CreateRenderPipeline([RenderPipelineDescriptor]).

Pipeline overridable constant ""enable_glyph_texture"" not found in [ShaderModule "wgpu/shader.wgsl"].
 - While validating vertex stage ([ShaderModule "wgpu/shader.wgsl"], entryPoint: "vs_main").
 - While validating vertex state.
 - While calling [Device].CreateRenderPipeline([RenderPipelineDescriptor]).

[Invalid RenderPipeline (unlabeled)] is invalid.
 - While encoding [RenderPassEncoder (unlabeled)].SetPipeline([Invalid RenderPipeline (unlabeled)]).
 - While finishing [CommandEncoder (unlabeled)].

[Invalid CommandBuffer] is invalid.
 - While calling [Queue].Submit([[Invalid CommandBuffer]])
```

OS: Windows
Browser: Chrome Version 131.0.6778.205

joshua-burbidge:
So I've spent some time looking into this.

I determined that the error I copied above can be avoided by adding a reference to each override constant in the vs_main entrypoint. like this:
```wgsl
@vertex
fn vs_main(
) -> VertexOutput {
    let _unused = shader_type;
    let _unused2 = enable_glyph_texture; 
    // the rest...
}
```

After that, I experience this error:
```
[...]

PR #234, opened 2025-03-15
Fix validation errors when running with WASM and wgpu feature

joshua-burbidge:
See issue #229 

This change prevents errors such as:
```
Pipeline overridable constant ""shader_type"" not found in [ShaderModule "wgpu/shader.wgsl"].
 - While validating vertex stage ([ShaderModule "wgpu/shader.wgsl"], entryPoint: "vs_main").
```
when running with WASM and the `wgpu` feature enabled.

----

Running example with:
```sh
cargo install wasm-bindgen-cli
cargo build --target=wasm32-unknown-unknown --example demo
wasm-bindgen ./target/wasm32-unknown-unknown/debug/examples/demo.wasm --out-dir examples/generated --target web

cd examples/
python3 -m http.server
```

Tested on Windows and Chrome 134.0.6998.89

Before: errors
image

After: success
image

### DIFF
--- a/src/renderer/wgpu.rs
+++ b/src/renderer/wgpu.rs
@@ -1188,17 +1188,10 @@ impl PipelineState {
         pipeline_layout: &wgpu::PipelineLayout,
         shader_module: &wgpu::ShaderModule,
     ) -> wgpu::RenderPipeline {
-        let constants = HashMap::from([
-            ("shader_type".to_string(), self.shader_type.to_f32() as f64),
-            (
-                "enable_glyph_texture".to_string(),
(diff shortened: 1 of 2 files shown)


joshua-burbidge:
@tronical I'm new to all this wgpu/graphics/shader stuff, so I'm not sure if this is a proper solution, but it resolves the error for me

tronical (approved):
Thanks for the analysis and fix. It somehow feels like a bug in wgpu, and it's a bummer that those fields aren't constants anymore. In the GLES code they were preprocessor defines and that made a big difference.

Anyway, I think getting things running on wasm is more valuable at this point. Thanks a lot for the fix!

PR #241, opened 2025-09-05
Add support for conic gradients

expenses:
I've tested both opengl and wgpu via slint. Happy to write a test here if needed.


tronical (approved):
Awesome :). The CI is nitpicking about formatting, but otherwise this LGTM.

(A naming thing I'm confused about is that I see folks referring to this as "conic gradients" as well as "conical gradients". In Slint we settled for the CSS `@conic-gradient`, but I don't mind either way here)

PR #246, opened 2025-10-07
Fix docs.rs build

nicoburns:
A recent change to Rust nightly means that doc_auto_cfg no longer builds. This affects the femtovg 0.19 release https://docs.rs/crate/femtovg/0.19.0.

The fix is to use `doc_cfg` instead of `doc_auto_cfg`


tronical (approved):
Thank you! I'll make a new upload tomorrow.

PR #271, opened 2026-04-13
External texture support and partial implementation in wgpu

expenses:
Required for https://github.com/slint-ui/slint/pull/10715

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1702,7 +1702,7 @@ impl Renderer for RecordingRenderer {
     ) -> Result<Self::Image, ErrorKind> {
         Err(ErrorKind::UnsupportedImageFormat)
     }
-    
+
     fn create_image_from_external_texture(
         &mut self,
         _external_texture: Self::ExternalTexture,
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1675,6 +1675,7 @@ pub struct RecordingRenderer {
 impl Renderer for RecordingRenderer {
     type Image = DummyImage;
     type NativeTexture = ();
+    type ExternalTexture = ();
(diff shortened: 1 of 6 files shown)


tronical (commented):
Looks correct. API wise I'm wondering though if perhaps this should be done with a separate function and type? The GL renderer can alias it to the same texture type, but with a separate function and type we can also support wgpu's external textures in the future.

Differently put: I think it makes sense to have this and do it this way, but I think with very little extra effort the API could be made future-proof so that we can support external wgpu textures as well in the future. So (1) separate function and (2) associated type for external texture.

What do you think?

expenses:
What do you mean by a seperate function? The main thing that'd need to be duplicated or changed with another `Renderer::Image` type is the render function. It's a lot of code duplication for not much benefit

expenses:
I think that `create_texture_view_and_sampler` needs to return a BindingResource and Sampler, where we have a separate bind group layout, bind group and pipeline for when we're using an external texture instead.

tronical:
> I think that `create_texture_view_and_sampler` needs to return a BindingResource and Sampler, where we have a separate bind group layout, bind group and pipeline for when we're using an external texture instead.

Sounds reasonable. I think it's okay indeed to do that later. This helps at least to get the API right. You could even make the implementation return an error if you'd like.

tronical:
Looks good. I think only two bits are missing:

1) the OES_EXTERNAL binding when the external texture is used, so that it works when sampling from it. (Like in your original commit)
2) the build failure in the CI

tronical (approved):
Thanks :). I think a small format fix is needed :(

PR #272, opened 2026-04-14
update to wgpu 29

Murmele:


Cargo.toml
tronical: I think you need to fix this line as well for the examples to compile (and get the CI green)

tronical (changes requested):
Thanks :)

Cargo.toml
Murmele: Oh thanks :)

Murmele:
```suggestion
wgpu = { version = "29" }
```

tronical:
#278 does the same but also fixes the build of the examples (which was missing here).

PR #278, opened 2026-05-13
Update wgpu backend to wgpu 29.0.3

matthargett:
I need this so I avoid linking two different versions of wgpu into my fork of BabylonNative that I've converted to use purely WebGPU / wgpu-native and eliminate GLES legacy code. I integration tested (screenshot attached) where I use JS to create a new texture via canvas, write text and boxes with gradients into the texture, and then BabylonJS renders that onto a spinning cube using the WebGPU JS API. I also tested with various sanitizers in this integration test as a precaution, even though Rust intrinsically prevents most of those issues.

all femtovg CI and unit tests pass in my femtovg fork, and it doesn't looks like it's a breaking change for existing users of femtovg. It would be awesome if you could merge and publish! :D

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,7 @@ image = { version = "0.25.0", optional = true, default-features = false }
 serde = { version = "1.0", optional = true, features = ["derive", "rc"] }
 glow = { version = "0.17.0", default-features = false }
 log = "0.4"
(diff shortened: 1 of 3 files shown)


tronical (approved):
Thank you :)

tronical:
Merged and published :)

PR #279, opened 2026-05-17
Add dashed stroke support

matthargett:
## Summary

- add line dash pattern and offset to `Paint`
- expand dashed strokes into visible path segments before stroke tessellation
- import SVG `stroke-dasharray` and `stroke-dashoffset` in the SVG example

References: follows SVG and Canvas 2D dash behavior for odd-length patterns and dash offsets. Path tests cover interval splitting, offsets, odd pattern repetition, and invalid zero patterns.

Split follow-ups: #280 for rounded scissor clips, #281 for glyph/text positioning.

## Validation

- `cargo fmt --check`
- `cargo test`

### DIFF
--- a/src/paint.rs
+++ b/src/paint.rs
@@ -918,19 +918,17 @@ impl Paint {
 
     /// Sets the dash pattern used for stroked paths.
     ///
-    /// Passing an empty slice, a pattern whose values sum to zero, or a pattern
-    /// containing non-finite or negative values resets strokes to solid. If the
-    /// supplied pattern has an odd number of entries, it is repeated once to
-    /// form an even pattern as required by SVG and Canvas 2D.
+    /// Passing an empty slice clears the dash pattern. Patterns containing
+    /// non-finite or negative values are ignored, leaving the current pattern
+    /// unchanged. If the supplied pattern has an odd number of entries, it is
+    /// repeated once to form an even pattern as required by SVG and Canvas 2D.
     pub fn set_line_dash(&mut self, dash: &[f32]) {
-        self.stroke.line_dash.clear();
-
-        if dash.is_empty() || dash.iter().any(|value| !value.is_finite() || *value < 0.0) {
+        if dash.iter().any(|value| !value.is_finite() || *value < 0.0) {
             return;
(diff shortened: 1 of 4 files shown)


tronical:
That’s quite a PR. Some things seem straightforward, otherwise not so much. Would you be willing to split this into smaller PRs?

I’m particularly curious about the changes to the glyph positioning.

matthargett:
> That’s quite a PR. Some things seem straightforward, otherwise not so much. Would you be willing to split this into smaller PRs?

yes, happy to do it. this was the smallest set of changes that go the screenshot comparison tests a LOT closer.

> I’m particularly curious about the changes to the glyph positioning.

Those changes definitely deserve their own review because they change when atlas text keeps fractional positions and when transformed text falls back to path rendering. The change did stop a weird bolding/smearing effect, and when I looked at over SVG implemtations the aligns closer to what they do.

I'll factor that out into its own PR as well and we can hash it out there.

thanks for taking the quick look! :D

matthargett:
Thanks, agreed. I split this up:

- this PR is now only dashed stroke support
- #280 has the rounded scissor implementation and tests
- #281 has the glyph/text positioning changes and a focused fractional-baseline test

The glyph positioning changes are no longer bundled with the dash/scissor work.

src/paint.rs
```diff
@@ -903,6 +907,67 @@ impl Paint {
...
+    pub fn line_dash(&self) -> &[f32] {
+        &self.stroke.line_dash
+    }
+
+    /// Sets the dash pattern used for stroked paths.
+    ///
+    /// Passing an empty slice, a pattern whose values sum to zero, or a pattern
+    /// containing non-finite or negative values resets strokes to solid. If the
+    /// supplied pattern has an odd number of entries, it is repeated once to
+    /// form an even pattern as required by SVG and Canvas 2D.
```
tronical: Canvas 2D seems to behave differently, or am I misreading this?

> If the number of elements in segments is odd, then let segments be the concatenation of two copies of segments.

tronical (approved):

PR #280, opened 2026-05-19
Add rounded scissor support

matthargett:
## Summary

- add `Canvas::rounded_scissor` and `Canvas::intersect_rounded_scissor`
- thread scissor radius through renderer params and uniforms
- apply the rounded scissor mask in OpenGL and wgpu shaders

References: follows the existing NanoVG/femtovg scissor-mask shader model for clipping, with Canvas-style rounded-rect clip use cases. Tests cover radius clamping and rectangular/rounded intersection behavior.

Split out of #279.

## Validation

- `cargo fmt --check`
- `cargo test`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1707,7 +1707,7 @@ where
             &glyph_texture,
             &scissor,
             1.0,
-            1.0,
+            self.fringe_width,
             -1.0,
         );
 
@@ -1954,6 +1954,19 @@ fn first_draw_params(commands: &[renderer::Command]) -> &Params {
         .expect("expected a draw command")
 }
 
+#[cfg(all(test, feature = "textlayout"))]
+fn first_glyph_draw_params(commands: &[renderer::Command]) -> &Params {
+    use renderer::CommandType;
+
+    commands
+        .iter()
+        .find_map(|command| match &command.cmd_type {
+            CommandType::Triangles { params } if params.glyph_texture_type != 0 => Some(params),
+            _ => None,
+        })
+        .expect("expected a glyph draw command")
+}
+
 #[cfg(test)]
 fn assert_approx_eq(actual: f32, expected: f32) {
     assert!(
@@ -1984,6 +1997,37 @@ fn rounded_scissor_radius_is_clamped_into_render_params() {
     let params = first_draw_params(&commands);
     assert_eq!(params.glyph_texture_type, 0);
     assert_approx_eq(params.scissor_radius, 10.0);
+}
+
+#[cfg(feature = "textlayout")]
+#[test]
+fn glyph_scissor_ramp_matches_fill_at_high_dpi() {
+    let renderer = RecordingRenderer::default();
+    let recorded_commands = renderer.last_commands.clone();
+    let mut canvas = Canvas::new(renderer).unwrap();
+    canvas.set_size(100, 100, 2.0);
+
+    let font = canvas
+        .add_font("examples/assets/RobotoFlex-VariableFont.ttf")
+        .expect("Font not found");
+    let paint = Paint::color(Color::white()).with_font(&[font]).with_font_size(16.0);
+
+    canvas.rounded_scissor(10.0, 10.0, 56.0, 28.0, 14.0);
+
+    let mut rect = Path::new();
+    rect.rect(0.0, 0.0, 100.0, 100.0);
+    canvas.fill_path(&rect, &Paint::color(Color::white()));
+    canvas.fill_text(12.0, 30.0, "Click", &paint).unwrap();
+    canvas.flush_to_output(());
+
+    let commands = recorded_commands.borrow();
+    let fill = first_draw_params(&commands);
+    let glyph = first_glyph_draw_params(&commands);
+
+    assert_approx_eq(glyph.scissor_scale[0], 2.0);
+    assert_approx_eq(glyph.scissor_scale[1], 2.0);
+    assert_approx_eq(glyph.scissor_scale[0], fill.scissor_scale[0]);
+    assert_approx_eq(glyph.scissor_scale[1], fill.scissor_scale[1]);
 }
 
 #[test]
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -1514,13 +1585,13 @@ where
...
-        // at a fixed size, so applying a non-translation transform to the textured quad
-        // produces blurry or aliased results.
+        // Very large glyphs need path rendering, and large transformed glyphs avoid
+        // magnifying atlas bitmaps. Smaller transformed text stays on the atlas so
+        // it keeps the browser-like hinted weight expected by canvas UI text.
         // Use 1e-3 as epsilon: tight enough to catch any intentional transform,
         // but generous enough to tolerate floating-point drift from matrix operations.
-        let need_direct_rendering = paint.text.font_size > 92.0 || !self.state().transform.is_pure_translation(1e-3);
+        let need_direct_rendering = paint.text.font_size > 92.0
+            || (paint.text.font_size > 64.0 && !self.state().transform.is_pure_translation(1e-3));
```
tronical: This looks like it's unrelated?

src/lib.rs
```diff
@@ -821,7 +839,60 @@ where
         let rect = Rect::new(tx - tex, ty - tey, tex * 2.0, tey * 2.0);
         let res = rect.intersect(Rect::new(x, y, w, h));
 
-        self.scissor(res.x, res.y, res.w, res.h);
+        let requested = Rect::new(x, y, w, h);
+        let tolerance = 1e-3;
```
tronical: Why did you choose 1e-3 here?

src/lib.rs
matthargett: Agreed. That text-rendering change leaked in from the split PR and does not belong here. I delete it from this PR branch.

src/lib.rs
matthargett: The `1e-3` was just intended to absorb transform/rectangle float drift as I tried to close the gap versus the reference render, but it was arbitrary. I replaced it with the existing canvas/path distance tolerance (`self.dist_tol`) so this uses the same geometry tolerance as the rest of femtovg. For rounded-corner containment I also changed the comparison to `(radius + tolerance)^2` so the units stay consistent.

tronical (approved):
Thanks :). Makes sense AFAICS

tronical:
@codex review

src/renderer/opengl/main-fs.glsl
```diff
@@ -50,6 +51,13 @@ float sdroundrect(vec2 pt, vec2 ext, float rad) {
 
 // Scissoring
 float scissorMask(vec2 p) {
+    if (scissorRadius > 0.0) {
+        vec2 pt = (scissorMat * vec3(p,1.0)).xy;
+        float distance = sdroundrect(pt, scissorExt, scissorRadius);
+        float edge = glyphTextureType != 0.0 ? 0.375 : 0.5;
```
tronical: This is one thing I haven't understood yet though: why do
Glyphs have a different behavior? How could this code even become
Active?

matthargett: it can become active for glyph atlas draws under a rounded scissor because glyph draw params set glyphTextureType non-zero. you're right, they should not behave differently. That `glyphTextureType != 0.0 ? 0.375 : 0.5` branch was a test-driven attempt to tighten rounded-scissor coverage for pre-antialiased glyph textures, where glyph alpha and clip alpha multiply together. (This scenario is in the BabylonNative multiple-gui regression test case.) But there is no good Canvas/SVG-style reason for the clip mask itself to depend on whether the draw is text. The clip should be draw-content agnostic.

This was me trying to reconcile more closely matching a text-heavy clip case and the "femto" part of the project name by keeping it .... 'concise', but the better fix is probably to keep rounded-scissor coverage content-agnostic and handle any remaining glyph mismatch in glyph positioning/coverage. I’ll try removing the glyph-specific edge and use the same 0.5 scissor edge for all draw types, and will try to expand the unit test and my local screenshot comparison cases to make this more generalized (while still respecting the `femto` name intent :D ).

tronical: > rounded-scissor coverage for pre-antialiased glyph textures

Oh is this for drawing glyphs (from the atlas) while a rounded scissor is set?

What if render text into an intermediate texture with other items and then draw the intermediate texture with rounded scissors?

matthargett: Yea, that working was referring to direct glyph atlas draws while a rounded scissor is active.

The intermediate-texture case is a good way to think about it tho: if text and other content are first rendered into a texture and then that texture is drawn under the same rounded scissor, it _should_ get the same clip coverage as direct glyph rendering. The earlier `glyphTextureType != 0.0 ? 0.375 : 0.5` branch would have made those two paths differ, which breaks down once I look at the broader WPT coverage for this area and beyond the BabylonNative screenshot mismatch I've been focused on.

That special case is removed in the latest commit. Rounded-scissor coverage is now content-agnostic again: glyphs, image/texture draws, and ordinary fills all use the same `0.5 - distance * scale` mask. The remaining glyph-related fix is only that the glyph draw path now passes the same device fringe width into the scissor params as fills/strokes, so at high DPI the scissor ramp scale is consistent across draw types (if the user of the library gives correct DPI, which BabylonNative doesn't, which is a me problem and not a femtovg problem).

tronical (approved):
Looks good to me - thanks :).

tronical:
Would you like to rebase and squash/tidy up the commits or would you prefer if I take them as-is? Or I could also squash.

matthargett:
> rebase and squash/tidy up the commi

you can squash, unless you want me to separate out the commits into specific diffs that highlight the teachable moments

tronical:
Oh oops m, it looks like there are conflicts - sorry about that:-/

matthargett:
> Oh oops m, it looks like there are conflicts - sorry about that:-/

I fixed those and pushed (and did the same for the other open PR as well). lmk if I'm missing anything, github UI shows everything green and good to go :D

PR #281, opened 2026-05-19
Preserve fractional text baseline in layout

matthargett:
Text layout rounded the baseline to an integer pixel while leaving the x position fractional, vertically snapping `measure_text` results and outline-rendered glyphs and dropping sub-pixel baseline precision.

Keep the baseline fractional, symmetric with x. This resolves text-positioning failures in BabylonNative's e2e screenshot suite, whose reference renders come from Chrome and NanoVG.

Net change over current master is the layout fix plus a fractional-baseline regression test.

## Validation
- `cargo fmt --check`
- `cargo test`

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -544,8 +544,12 @@ impl GlyphAtlas {
 
                 let line_width_offset = if rendered.color_glyph { 0. } else { line_width_offset };
 
-                q.x0 = glyph.x.trunc() + rendered.bearing_x as f32 - line_width_offset - GLYPH_PADDING as f32;
-                q.y0 = glyph.y.round() - rendered.bearing_y as f32 - line_width_offset - GLYPH_PADDING as f32;
+                let transformed_text = !canvas.state().transform.is_pure_translation(1e-3);
+                let glyph_x = if transformed_text { glyph.x } else { glyph.x.trunc() };
+                let glyph_y = if transformed_text { glyph.y } else { glyph.y.round() };
+
+                q.x0 = glyph_x + rendered.bearing_x as f32 - line_width_offset - GLYPH_PADDING as f32;
+                q.y0 = glyph_y - rendered.bearing_y as f32 - line_width_offset - GLYPH_PADDING as f32;
                 q.x1 = q.x0 + rendered.width as f32;
                 q.y1 = q.y0 + rendered.height as f32;
 
--- a/src/text.rs
+++ b/src/text.rs
@@ -434,6 +434,12 @@ pub struct GlyphDrawCommands {
     pub color_glyphs: Vec<DrawCommand>,
 }
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum AtlasGlyphPositioning {
+    PixelAligned,
+    Subpixel,
+}
+
 pub struct GlyphAtlas {
     pub rendered_glyphs: RefCell<FnvHashMap<RenderedGlyphId, Option<RenderedGlyph>>>,
     pub glyph_textures: RefCell<Vec<FontTexture>>,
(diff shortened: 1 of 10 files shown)


src/text.rs
```diff
@@ -544,8 +544,12 @@ impl GlyphAtlas {
 
                 let line_width_offset = if rendered.color_glyph { 0. } else { line_width_offset };
 
-                q.x0 = glyph.x.trunc() + rendered.bearing_x as f32 - line_width_offset - GLYPH_PADDING as f32;
-                q.y0 = glyph.y.round() - rendered.bearing_y as f32 - line_width_offset - GLYPH_PADDING as f32;
+                let transformed_text = !canvas.state().transform.is_pure_translation(1e-3);
```
tronical: This is a condition I'd like to adjust soon, to also support uniform scaling. It would be nice to pass is through and make the decision in once place in draw_glyph_run(). What do you think?

src/text.rs
```diff
@@ -544,8 +544,12 @@ impl GlyphAtlas {
...
+                let glyph_x = if transformed_text { glyph.x } else { glyph.x.trunc() };
```
tronical: IMO this deserves a comment - this code is already rather magic :)

src/text/textlayout.rs
```diff
@@ -542,7 +542,7 @@ fn layout(
 
     for glyph in &mut res.glyphs {
         glyph.x = cursor_x + glyph.offset_x;
-        glyph.y = (cursor_y + alignment_offset_y).round() + glyph.offset_y;
+        glyph.y = cursor_y + alignment_offset_y + glyph.offset_y;
```
tronical, line 545: Why do you remove the rounding here? Don't we want to pixel align on the y coordinate if possible?

src/text.rs
matthargett: It makes sense to collate this responsibility, for sure. I changed this so `draw_glyph_run()` computes the transform classification once, creates an atlas positioning policy, and passes that into `render_atlas()`. That should keep the future uniform-scaling decision centralized.

src/text.rs
matthargett: I added a comment around the atlas quad placement.

My intent is that layout should preserve the logical fractional baseline, and pixel alignment should happen at atlas draw time only when it is safe. For pure translations we still truncate/round the atlas quad position like before. For transformed text, local-space rounding becomes visible after the transform, which was the source of the smeared/offset text versus browser/reference image mismatch this PR is trying to reduce.

That also matches what I see in browser engines at this layer: Canvas text carries float positions into the text stack, and rasterization/hinting decisions are made lower down rather than by rounding the layout baseline up front. WebKit carries FloatPoint through canvas text and FontCascade::drawGlyphBuffer; Firefox carries gfx::Point into gfxTextRun::Draw.

src/text/textlayout.rs
matthargett: great question! I wouldn’t frame this as “never pixel align.” The follow-up commit keeps pixel alignment for pure-translation atlas text, but preserves fractional layout positions so transformed text can avoid rounding in local space before the transform is applied.

That split matches the browser-shaped model better: WebKit/Firefox carry float canvas text positions into their text drawing stacks, with rasterization/hinting handled lower down. The BabylonNative screenshot was the motivation, but the intended renderer rule is: preserve layout positions, then pixel-align atlas quads only when the transform makes that safe. (“safe” for me means: doesn’t propagate the precision loss of the rounding/truncation or change the intended visual position after the transform.)

btw, I’m not trying to claim femtovg needs to match every browser backend detail, just that preserving the fractional layout position and deciding pixel alignment at draw time seems like the right split in my head and looking at other reference implementations reinforced that.

matthargett:
I see your commit that fixes this in a better way. I'll narrow this PR to the float rounding issues

matthargett:
@tronical anything you'd like me to do/change in this PR? I re-scope it and rebased based on your merges that took a different approach.

tronical (approved):
Thank you so much for your patience and minimising the patch. I agree with it :).

PR #285, opened 2026-06-14
Add Canvas 2D drop shadows (shadowColor / shadowBlur / shadowOffset)

matthargett:
Adds Canvas 2D drop shadows — `shadowColor`, `shadowBlur`, `shadowOffsetX`, `shadowOffsetY` (via `set_shadow_color` / `set_shadow_blur` / `set_shadow_offset`, stacked by `save()`/`restore()`) — for fills, strokes, and text. Reuses the existing Gaussian image-filter and render-to-texture machinery; no new dependencies and no new shader entry points.

When the shadow color is non-transparent, each draw renders the shape's coverage into a transient offscreen image, recolors it by the shadow color preserving the source's alpha, Gaussian-blurs it (σ = `shadowBlur`/2 per the HTML drawing model), and composites it under the shape at the offset. The default (transparent shadow color) adds no new work.

### Conformance (WHATWG Canvas drawing model + WPT `2d.shadow.*`)
- Defaults `rgba(0,0,0,0)` / `0` / `(0,0)`; a shadow is drawn only when the color is non-transparent **and** one of blur/offsetX/offsetY is non-zero.
- Shadow offset is in device space (not transformed by the CTM).
- The shadow is built from the **source's alpha**, so a semi-transparent fill / gradient / image-with-holes casts a proportionally weaker shadow and a fully transparent source casts none.
- `shadowColor` alpha and `globalAlpha` both modulate the shadow.
- Non-finite blur/offset values are ignored (setter semantics).
- cross-checked against Firefox and WebKit's non-WPT test suite to make sure I wasn't repeating mistakes, encoded those lessons as inline unit tests

### Blur-falloff adjustment (also corrects the shared Gaussian blur)
While validating against Chrome Canary I found the Gaussian blur was systematically **~21% too tight**: the shader sampled only `±1.5σ` taps and renormalized by that partial sum, shrinking the effective σ to ~0.79× the spec `σ = blur/2`. The offscreen source was already padded to 3σ and the docs already claimed 3σ reach — only the shader tap count disagreed. This PR widens the sampled reach to `±3σ` and raises the GLES 2.0 constant loop bound from 12 to 24 so 3σ is covered at the σ≤8 cap; the per-tap weights were already the true-σ Gaussian, so this only restores the missing tail, and the early `break` keeps small blurs cheap.

Because `renderFilteredImage` is shared, **this also fixes the existing `ImageFilter::GaussianBlur` image-blur filter** (which was equally too tight). I'm happy to split the blur-kernel change into its own PR and rebase this on top if you'd prefer to keep this one shadow-only.

Verified against Chrome Canary 151 (headless, dpr=1): effective σ now matches at `shadowBlur` 12 and 16 (1.00×, was 0.79×), and the full shadow scene drops to a max per-channel difference of **2/255**. Very large `shadowBlur` (>16) is reach-capped by the 24-tap bound (~94% of target σ at blur 24); `shadowBlur` ≤ 16 is exact.

femtovg vs Chrome Canary — drop shadow

*Left → right: Chrome Canary, femtovg, ×4-amplified difference.*

### Tests
- State / command-inspection (no GPU needed): attribute defaults, the blur-or-offset enable rule, non-finite rejection, save/restore stacking, and transparent-color → no shadow pass.
- wgpu-gated headless render tests (skip gracefully without an adapter): offset placement, offset-not-scaled-by-CTM, source-alpha attenuation, off-screen-shape-with-offset still casts, and a blur-falloff width regression test that fails on the old `±1.5σ` kernel that failed screenshot comparison and passes with the fix.

### Validation
`cargo fmt --check`, `cargo build`, `cargo build --features wgpu`, `cargo test --features "wgpu textlayout"`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -522,6 +522,19 @@ where
     /// shadow color is fully transparent no offscreen shadow pass is performed and
     /// drawing adds zero overhead. The shadow color's own alpha multiplies the
     /// shadow's coverage.
+    ///
+    /// # Performance
+    ///
+    /// Shadows are not cheap: every shadowed fill, stroke or text draw renders the
+    /// shape's coverage into a transient offscreen image sized to its padded
+    /// bounds, runs a two-pass Gaussian blur over it (when `shadowBlur` is
+    /// non-zero), and composites the result — per draw, every frame. Prefer
+    /// shadowing a few composed shapes over many small primitives, and if the
+    /// same shadowed shape is drawn every frame, consider rendering it once into
+    /// an [image](Self::create_image_empty) via
+    /// [`set_render_target`](Self::set_render_target) and re-drawing that cached
+    /// image instead. Setting a fully transparent shadow color restores the
+    /// zero-overhead path.
     pub fn set_shadow_color(&mut self, color: Color) {
         self.state_mut().shadow_color = color;
     }
@@ -1455,10 +1468,18 @@ where
         else {
             return;
         };
-        let Ok(blurred_image) = self.create_image_empty(width, height, PixelFormat::Rgba8, ImageFlags::PREMULTIPLIED)
-        else {
-            self.delete_image(coverage_image);
-            return;
+        // The blur kernel divides by sigma, so a zero (or sub-pixel) blur skips
+        // the filter pass entirely — and with it the second offscreen image.
+        let blurred_image = if sigma >= 0.01 {
+            match self.create_image_empty(width, height, PixelFormat::Rgba8, ImageFlags::PREMULTIPLIED) {
+                Ok(image) => Some(image),
+                Err(_) => {
+                    self.delete_image(coverage_image);
+                    return;
+                }
+            }
+        } else {
+            None
         };
 
         let previous_target = self.current_render_target;
@@ -1505,10 +1526,9 @@ where
 
         self.restore();
 
-        // Blur the coverage into the second offscreen image. The blur kernel
-        // divides by sigma, so for a zero (or sub-pixel) blur we skip the filter
-        // entirely and composite the sharp coverage directly.
-        let source_image = if sigma >= 0.01 {
+        // Blur the coverage into the second offscreen image; a sharp shadow (no
+        // blur image allocated) composites the coverage directly.
+        let source_image = if let Some(blurred_image) = blurred_image {
             self.filter_image(blurred_image, ImageFilter::GaussianBlur { sigma }, coverage_image);
             blurred_image
         } else {
@@ -1550,7 +1570,9 @@ where
         // The transient images are referenced by deferred draw commands, so they
         // can only be freed after the next flush. Queue them for later cleanup.
         self.shadow_images.push(coverage_image);
-        self.shadow_images.push(blurred_image);
+        if let Some(blurred_image) = blurred_image {
+            self.shadow_images.push(blurred_image);
+        }
     }
 
     /// Frees offscreen images allocated by drop-shadow passes during the frame.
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1556,6 +1883,44 @@ where
             text::normalize_variations(&text_context, &paint.text.font_ids, &paint.text.font_variations)
         };
 
+        // Draw the drop shadow (if any) under the text. The shaped layout gives a
+        // user-space box; transform its corners by the CTM to obtain device-space
+        // bounds and let render_shadow re-enter draw_text with the shadow tint.
+        // (render_shadow disables shadows in the state so this does not recurse.)
+        if self.shadow_enabled() {
```
tronical, line 1922: This makes sense, to draw the shadow first. But I think there's a bug: If the text rendering hits the path of drawing the outlines, it'll draw each with the shadow as well, won't it? I think shadow_enabled() must return false for the rest of the function after that.

src/lib.rs
```diff
@@ -1234,6 +1349,218 @@ where
...
+        // images as PREMULTIPLIED. Otherwise the image-sampling shader would
+        // re-premultiply on composite (multiplying rgb by alpha a second time),
+        // darkening partially-transparent shadow texels — which the source-alpha
+        // shadow now produces wherever the source is semi-transparent or
+        // antialiased.
+        let Ok(coverage_image) = self.create_image_empty(width, height, PixelFormat::Rgba8, ImageFlags::PREMULTIPLIED)
+        else {
+            return;
+        };
+        let Ok(blurred_image) = self.create_image_empty(width, height, PixelFormat::Rgba8, ImageFlags::PREMULTIPLIED)
```
tronical: This could be protected with `if sigma >= 0.01` to avoid an unnecessary allocation.

tronical (commented):
This looks great overall - thanks! I think there's a bug with the text rendering (see comment inline). I think the only concern I have here is that when a shadow is enabled, the price users pay is _a lot_ for each frame. Perhaps this should be documented?

Apart for the minor notes I think this is good to have :)

As a side-note: I added the original filter api so that we can do box shadows in Slint with femtovg, which also works well (easy shape...). But in order for that to not ruin rendering times, we cache the box shadows.

src/lib.rs
matthargett: Good catch — confirmed real. Glyph runs that fall back to outline rendering go through `fill_path_internal`/`stroke_path_internal`, whose own shadow hooks fired again per glyph on top of the run-level shadow: double-darkened text and one offscreen pass *per glyph*. Fixed in b76845d  as  requested: once the run-level shadow is drawn, shadows are suppressed for the remainder of `draw_text` (restored on error paths too). Locked with a regression test that asserts the offscreen-pass count is glyph-count-invariant — before the fix it fails at 2 passes for "I" vs 11 for "Illuminate".

I'm cross-checking in WPT now to make sure this doesn't introduce some unintended conformance consequence

src/lib.rs
matthargett: Done in https://github.com/femtovg/femtovg/pull/285/commits/209c6d23b186e3103e638a9aa60fbce0151f9c25 — the second offscreen image is now only allocated when `sigma >= 0.01`; sharp shadows composite the coverage image directly with no extra allocation.

matthargett:
Thanks for the review! All three points addressed in b76845d:

- **Text double-shadow**: real bug, fixed as you suggested — shadows are suppressed for the rest of `draw_text` after the run-level shadow, with a glyph-count-invariance regression test (2 vs 11 offscreen passes before the fix).
- **`sigma >= 0.01` allocation guard**: done; sharp shadows no longer allocate the blur image.
- **Per-frame cost**: added a `# Performance` section on `set_shadow_color` documenting that every shadowed draw is an offscreen coverage pass + two-pass blur + composite, and recommending shadowing a few composed shapes and/or caching the result into an image via `set_render_target` for anything redrawn every frame — the same strategy as your cached box shadows in Slint. A built-in cache (keyed on shape + shadow params) would be a natural follow-up if there's appetite.

Also adapted the shadow GPU test harness to the wgpu 30 bump that landed on the branch via the master merge (`RequestAdapterOptions` gained a field; `get_mapped_range` now returns a `Result`) — the full suite is green on both backends again.

This is deep enough that I'm dong an extra fuzz and branch mutation pass to try and fortify against other corner cases. this is trickier to cover in a durable way than it first appeared!

matthargett:
Two more fixes pushed in b03105a, both surfaced by tightening the text-shadow tests into a spec-derived pixel assertion (shadow alpha must never exceed shadowColor's alpha, per the drawing model's union-alpha rule — checked on both the atlas and outline rasterization paths, with overlapping glyphs via negative letter spacing):

- **Shadows were composited vertically mirrored.** Image render targets store their content flipped in texture v-space (the FBO convention both backends share); the glyph atlas and `filter_image` each compensate, but `render_shadow` composited its coverage assuming upright storage, mirroring every shadow about its padded box's midline. All earlier shadow tests used mirror-symmetric shapes (rects/circles), so it was invisible; a tall text run mirrored its shadow fully off-canvas. Fixed by creating the offscreen images with `ImageFlags::FLIP_Y`, which the shared paint-matrix code already honors on both backends. A positional probe now confirms shadows land at exactly ink + offset.
- **Text-run shadow bounds** now derive from the union of the glyph boxes rather than the advance-summed layout width, which collapses to zero under strongly negative letter spacing while ink is still painted.

Full suite is green on both backends (73 tests), and the new union-alpha test fails on either defect if reintroduced.

matthargett:
@tronical anything other testing or changes you wanted me to do?

tronical (approved):
Thank you :)

PR #290, opened 2026-07-04
Add SVG elliptical arc path builder (Path::svg_arc_to)

matthargett:
Adds `Path::svg_arc_to(rx, ry, x_axis_rotation, large_arc, sweep, x, y)` — the endpoint parameterization of the SVG path `A` command — emitting cubic béziers through the existing tessellation pipeline. Pure path-building code: no shader/uniform surface, no new dependencies.

Previously `arc()`/`arc_to()` were circular-only, so consumers of SVG path data (or Canvas `Path2D` arcs) had to re-derive the endpoint→center conversion themselves.

Implements W3C SVG 1.1 §F.6: endpoint→center conversion (F.6.5), out-of-range radii correction (F.6.6), the degenerate rules (F.6.2), and the F.6.5.4 sign convention at the 180° boundary (a zero cross product takes the positive angle). Non-finite arguments are rejected with an early return, matching the Canvas rule for path methods. The center parameterization is computed in f64 (as in resvg/kurbo/Batik): fuzzing found two f32 overflow cases — spec-valid tiny radii (F.6.6 scale-up of rx=ry=1e-30 across a 100-unit chord) and huge endpoint coordinates (±1e30) — that emitted NaN control points; both are fixed and kept as regression tests, with a finite-output chord fallback for geometry unrepresentable even in f64.

### Validation
- **Pixel comparison vs Chrome Canary** (`Path2D('A 100 60 30 1 0 …')`, headless, dpr=1) plus a 12-case sweep over all large-arc/sweep combinations, rotated and flat ellipses, radius scaling, semicircles, and negative rotation: zero true coverage-gap pixels; only stroke-AA fringe differs.

  arc comparison

- **Deterministic property fuzz**: 20,000 seeded iterations (coords ±1e4, radii 0–2e3, rotations ±10 rad, all flags, with extreme-magnitude injections) asserting no panics, finite output, endpoint reached, sampled points on the rotated ellipse, and sweep direction/magnitude consistent with the flags. Runs in ~0.06 s as a normal unit test.
- **Mutation pass**: 7 hand-applied mutations of the F.6 math (sign rule, radius scaling, rotation, sweep-modulo, kappa handles, angle sign, rx/ry swap) — all killed by the committed test suite. The one initial survivor (swapped sweep-modulo, which stays on the correct ellipse but sweeps the complementary direction with an extra revolution) is now killed by a parametric-angle-unwrapping test covering all four flag combinations.
- **Cross-surface isolation**: tests cover one `Path` shared by two canvases with different transforms (interleaved, byte-identical per-canvas output), render-target interleaving (Image/Screen), and multi-threaded per-thread usage; a `compile_fail` doctest documents that `Path` is intentionally `!Sync`.

### Checks
- `cargo fmt --check`
- `cargo test --features "wgpu textlayout"` (77 tests)

Rebased onto current master (which now contains #285 / #292 / #294); this PR is `src/path.rs` + tests only, no overlap with those. Ready for review.

### DIFF
--- a/src/path.rs
+++ b/src/path.rs
@@ -1510,6 +1510,236 @@ mod tests {
         }
     }
 
+    // ---- deterministic property fuzz ----
+
+    /// Minimal deterministic PRNG (Knuth's MMIX LCG constants, high 32 bits
+    /// used) so the fuzz below needs no external crates and replays
+    /// identically on every run.
+    struct Lcg(u64);
+
+    impl Lcg {
+        fn next_u32(&mut self) -> u32 {
+            self.0 = self
+                .0
+                .wrapping_mul(6364136223846793005)
+                .wrapping_add(1442695040888963407);
+            (self.0 >> 32) as u32
+        }
+
+        /// Uniform in [0, 1).
+        fn unit(&mut self) -> f32 {
+            (self.next_u32() >> 8) as f32 / (1u32 << 24) as f32
+        }
+
+        fn range(&mut self, lo: f32, hi: f32) -> f32 {
+            lo + (hi - lo) * self.unit()
+        }
+
+        fn flag(&mut self) -> bool {
+            self.next_u32() & 1 == 1
+        }
+
+        fn one_in(&mut self, n: u32) -> bool {
+            self.next_u32() % n == 0
+        }
+    }
+
+    fn fuzz_coord(rng: &mut Lcg, extreme: bool) -> f32 {
+        if extreme && rng.one_in(3) {
+            let magnitude = if rng.flag() { 1e30 } else { 1e-30 };
+            if rng.flag() {
+                magnitude
+            } else {
+                -magnitude
+            }
+        } else {
+            rng.range(-1e4, 1e4)
+        }
+    }
+
+    fn fuzz_radius(rng: &mut Lcg, extreme: bool) -> f32 {
+        if extreme && rng.one_in(3) {
+            if rng.flag() {
+                1e30
+            } else {
+                1e-30
+            }
+        } else if rng.one_in(40) {
+            0.0
+        } else {
+            rng.range(0.0, 2e3)
+        }
+    }
+
+    /// Independent f64 evaluation of the F.6.5/F.6.6 center parametrization,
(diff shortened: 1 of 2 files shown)


src/path.rs
```diff
@@ -414,6 +426,190 @@ impl Path {
...
+        // The sweep never exceeds a full turn (|delta| <= 2*pi), so at most four
+        // 90-degree segments are ever emitted; the extra headroom only guards
+        // against floating-point spill past the exact quarter-turn boundaries.
+        const MAX_SEGMENTS: usize = 8;
+        let ndivs = ((delta.abs() / (std::f64::consts::PI * 0.5)).ceil() as usize).clamp(1, MAX_SEGMENTS);
+        let seg = delta / ndivs as f64;
+        // Maisonobe per-segment handle length matching the kappa technique used
+        // by arc()/ellipse(): alpha = sin(seg) * (sqrt(4 + 3 tan(seg/2)^2) - 1) / 3.
+        let half = seg * 0.5;
+        let alpha = seg.sin() * ((4.0 + 3.0 * half.tan() * half.tan()).sqrt() - 1.0) / 3.0;
```
tronical: The comments above says that it's the kappa used by arc(), but this looks quite different. Is that intentional?

matthargett: besides the comment drifting away from the code (fixed and pushed), I was re-reading the tests and noticed that ELLIPSE_RESIDUAL_TOL was 5e-3 on the squared residual, which was way too loose. once I tightened to 1e-3 locally six of them failed. That expression was a different approximation from arc()'s kappa, and a way less accurate one: sampling the emitted curves against the true ellipse, worst-case relative radial error was pretty bad but didn't manifest in my screenshot testing. 

pushed a commit so the builders now genuinely agree. The tangent form also carries the sign of the sweep, dropping a branch so keeping with the 'femto' theme :) I'm eyeballing more screenshot tests that use this feature to try and see if there's any other subtle math problems that could yield a visual difference.

tronical (approved):
I'd say the tests are a tad bit too excessive, but otherwise I suppose this can't really break anything :)

PR #291, opened 2026-07-10
Add text decoration (underline / strikethrough / overline)

matthargett:
Adds underline, strikethrough and overline for `fill_text`/`stroke_text`, filling the long-standing `FontMetrics` TODO.

**Rebased onto current master** (drop shadows #285 landed). The two features interact in `draw_text`; the interaction fixes below are now folded into this PR and covered by a wgpu render test. Ready for review.

`FontMetrics` gains underline/strikeout position + thickness (from ttf-parser's `post`/OS-2 tables, with fallbacks when a font omits them), populated consistently in both the ttf-parser and swash metric paths. A `TextDecoration { underline, strikethrough, overline }` set on the text `Paint` draws the enabled lines in user space through the normal fill pipeline — the same space and CTM as the glyph quads, so lines stay aligned on the pure-translation atlas, scale-baked atlas, and outline rasterization paths alike. All enabled lines of a run are emitted as **one** multi-contour path and **one** fill (a single allocation set per run; `NonZero` forced so clamped thin bands cannot cancel), asserted strictly in the tests.

Canvas 2D has no text-decoration, so this is an extension toward SVG/CSS `text-decoration` parity (position/thickness from font metrics; the decoration takes the text fill).

### Comparison vs Chromium Canary (same font)

CSS `text-decoration` on the same RobotoFlex font; the meaningful comparison is decoration-line placement, which both derive from the font tables:

text-decoration comparison

With drop shadows (from the #285 branch), each decoration × sharp / blurred / colored shadows — one shadow covers glyphs and lines together, matching `filter: drop-shadow` semantics over a decorated element:

text-decoration with drop shadows

### Interaction with drop shadows (integrated in this PR)

Cross-testing against drop shadows found two interaction bugs, both fixed here and covered by the `decoration_shares_the_single_text_shadow` wgpu test (each decoration must be covered by exactly one run shadow — not doubled, not missing):

- decorations drew after the run shadow's state was restored, so each line cast a second shadow on top of the run's (compounded alpha);
- the shadow coverage bounds came from the glyph ink box, which is too short to reach an overline at the ascent for x-height-only runs, silently clipping its shadow — the bounds now use the font line box.

Both fixes are in this PR: the `draw_text` shadow pass covers glyphs + decorations, and decorations are drawn while shadows are suppressed so they share the single run shadow.

### Known differences, deliberately out of scope

- **`text-decoration-skip-ink`**: Chromium (and Firefox) break the underline around descenders (`g`, `y`) by default; SVG draws a continuous line, and that is the default here. Ink skipping needs glyph-outline/band intersection and is planned as a separate opt-in follow-up.
- **Overline position**: fonts carry no dedicated overline metric; the line sits at the ascent, which places it slightly higher than Chromium's choice. Cosmetic; tracked as a follow-up.

### Tests

Metrics report non-zero underline/strikeout for a known font; exactly one decoration fill per run with rects at the metric-derived y/thickness (none when disabled); alignment under a 2× scaled-atlas transform; graceful fallback for a font stripped of `post`/OS-2; serde round-trip and back-compat (`serde(default)`).

### Validation

- `cargo fmt --check`
- `cargo test --features "wgpu textlayout"` and `cargo test --features "serde wgpu textlayout"`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,6 +60,7 @@ use geometry::*;
 
 mod paint;
 pub use paint::Paint;
+pub use paint::TextDecoration;
 use paint::{GlyphTexture, PaintFlavor, StrokeSettings};
 
 mod path;
@@ -1923,27 +1924,55 @@ where
             text::normalize_variations(&text_context, &paint.text.font_ids, &paint.text.font_variations)
         };
 
+        // Captured in scaled shaping space; decorations hang off it below.
+        let baseline_scaled = run_baseline(&layout.glyphs);
+
         // Draw the drop shadow (if any) under the text. The shaped layout gives a
         // user-space box; transform its corners by the CTM to obtain device-space
         // bounds and let render_shadow re-enter draw_text with the shadow tint.
-        // (render_shadow disables shadows in the state so this does not recurse.)
+        // The re-entered draw_text draws glyphs *and* decoration lines (with
+        // shadows suppressed), so a single shadow covers the whole painting
+        // operation, matching the drawing model — decorations are not shadowed
+        // separately. (render_shadow disables shadows in the state so this does
+        // not recurse.)
         if self.shadow_enabled() {
             // Layout metrics are in the scaled shaping space; bring them back to
-            // user space. The horizontal extent is derived from the union of the
-            // glyph boxes rather than the advance-summed layout width: negative
-            // letter spacing can collapse the summed width to zero while ink is
-            // still painted, and the shadow must cover everything that is drawn.
-            // Expand by the line height on all sides so bearings, overhang,
-            // ascenders, descenders and diacritics are fully covered.
+            // user space. The horizontal extent unions the glyph ink boxes with
+            // the run's advance box (layout.x .. layout.x + width): negative letter
+            // spacing can collapse the advance width while ink is still painted,
+            // yet the decoration lines are drawn across that advance box, so the
+            // shadow must cover both.
             let (mut gx0, mut gx1) = (f32::INFINITY, f32::NEG_INFINITY);
             for glyph in &layout.glyphs {
                 gx0 = gx0.min(glyph.x);
                 gx1 = gx1.max(glyph.x + glyph.width);
             }
-            let ly = layout.y * invscale;
-            let lh = layout.height() * invscale;
             let (ux0, uy0, ux1, uy1) = if gx0 <= gx1 {
-                (gx0 * invscale - lh, ly - lh, gx1 * invscale + lh, ly + lh + lh)
+                let rx0 = layout.x.min(layout.x + layout.width());
+                let rx1 = layout.x.max(layout.x + layout.width());
+                let x0 = (gx0 * invscale).min(rx0 * invscale);
+                let x1 = (gx1 * invscale).max(rx1 * invscale);
+                // The vertical extent is the font's line box (baseline - ascent ..
+                // baseline - descent), not the glyph ink box: the overline sits at
+                // the ascent and the underline below the baseline, so an ink-box
+                // extent clips the decoration shadows of short (x-height-only)
+                // runs like "www". Deriving it from the metrics covers glyphs,
+                // decorations, ascenders, descenders and diacritics alike.
+                let baseline = baseline_scaled.map_or(layout.y * invscale, |b| b * invscale);
+                let (ascent, descent) = {
+                    let text_context = self.text_context.borrow();
+                    text_context
+                        .measure_font(paint.text.font_size, &paint.text.font_ids, &paint.text.font_variations)
+                        .map_or((0.0, 0.0), |m| (m.ascender(), m.descender()))
+                };
+                // A little slack for the overline nudge and antialiased edges.
+                let margin = paint.text.font_size * 0.2;
+                (
+                    x0 - margin,
+                    baseline - ascent - margin,
+                    x1 + margin,
+                    baseline - descent + margin,
+                )
             } else {
                 // No drawable glyphs: nothing painted, nothing to shadow.
                 (0.0, 0.0, 0.0, 0.0)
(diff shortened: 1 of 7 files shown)


matthargett:
@tronical anything I can do to help move the text PRs forward? looks like @RobertMueller2 is ready to adopt into their project whenever the feature gets released :)

src/lib.rs
```diff
@@ -2004,14 +2052,95 @@ where
...
+        // glyph runs use. That keeps the lines aligned with the glyphs across the
+        // direct-outline, atlas, and scale-baked-atlas paths alike.
+        //
+        // They are drawn while shadows are still suppressed: the run-level shadow
+        // pass above already rendered the decorations into its coverage (it
+        // re-enters draw_text, which draws them), so glyphs and decorations share
+        // a single shadow and the lines must not cast a second one.
+        if glyph_run_result.is_ok() && !paint.text.text_decoration.is_none() {
+            if let Some(baseline_scaled) = baseline_scaled {
+                self.draw_text_decorations(paint, baseline_scaled * invscale, layout.x, layout.width());
```
tronical: I think that this might be passing the wrong width though. Doesn't `layout.width()` include the advance of the last glyph, instead of the actual width?

src/lib.rs
```diff
@@ -3224,3 +3356,480 @@ fn outline_text_shadow_pass_count_is_glyph_count_invariant() {
...
+        let (mut canvas, commands, verts, font) = make_canvas();
+        canvas
+            .fill_text(50.0, baseline_y, "Hello", &base_paint().with_font(&[font]))
+            .unwrap();
+        canvas.flush_to_output(());
+        let fills = recorded_decoration_fills(&commands.borrow(), &verts.borrow());
+        assert!(fills.is_empty(), "expected no decoration fill, got {fills:?}");
+    }
+
+    // Underline: one fill command with one rect, centered below the baseline at
```
tronical, line 3594: The tests here seem to check only the y position of the decoration. It would be great if these could also check that the decorations are of the correct width and at the right position. Think of alignment or RTL affecting these and the value test coverage would add :)

src/lib.rs
```diff
@@ -2004,14 +2052,95 @@ where
...
+        // A vertical writing mode (top-to-bottom) would need the lines to run
+        // along y and offset along x, driven by vertical metrics; the geometry
+        // below would have to be generalized to the advance axis rather than
+        // hardcoding x as the run axis.
+        let decoration = paint.text.text_decoration;
+
+        // Metrics in user-space units (scaled for the unscaled font size) so the
+        // offsets match the user-space baseline. `measure_font` reads the primary
+        // font with the same variations the run was shaped with.
+        let metrics = {
```
tronical: The fetching of the metrics and the computation of the decoration extent (width/height) is duplicated here and earlier in the `if shadow()` branch. Those are tricky, I think it would be much better if those were  centralised.

src/paint.rs
```diff
@@ -1122,6 +1152,45 @@ impl Paint {
...
+    /// Returns the paint with the text decoration set to the specified value.
+    #[inline]
+    pub fn with_text_decoration(mut self, decoration: TextDecoration) -> Self {
+        self.set_text_decoration(decoration);
+        self
+    }
+
+    /// Returns the paint with underline/strikethrough/overline toggled.
+    ///
+    /// A convenience wrapper over [`with_text_decoration`](Self::with_text_decoration).
```
tronical: I'd leave out the convenience wrapper for what I think isn't _that_ critical API that requires convenience ;)

src/lib.rs
```diff
@@ -328,6 +329,25 @@ pub struct Canvas<T: Renderer> {
...
+/// `layout` positions each glyph as
+/// `glyph.y = cursor_y + alignment_offset + glyph.offset_y`, so every glyph
+/// shares the same (possibly fractional) `cursor_y + alignment_offset` baseline
+/// and the per-glyph term is `glyph.offset_y` (the GPOS y-offset). Recovering the
+/// baseline as `glyph.y - glyph.offset_y` is what keeps text decorations anchored
+/// to the run baseline even when the first drawable glyph carries a non-zero
+/// offset (e.g. text starting with a combining mark), instead of riding that mark
+/// up or down.
+#[cfg(feature = "textlayout")]
+fn run_baseline(glyphs: &[text::ShapedGlyph]) -> Option<f32> {
```
tronical: Instead of computing the baseline again, wouldn't it perhaps be easier to store that in TextMetrics in `layout()`?

tronical (commented):
Looks overall great - thanks! A few points inline that should - I think - be relatively straightforward to fix.

src/lib.rs
matthargett: in my reading and conceptual understanding from the spec, layout.width() is the advance sum, so it _intentionally_ does include the last glyph's advance. Since Canvas 2D has no text-decoration, SVG text-decoration is the reference implementation here, and browsers decorate the advance box, not the ink extent: rendering <text text-decoration="underline">Hi&#160;</text> in Chromium, the underline spans exactly [x, x + advance] and extends past the last inked pixel to cover the trailing no-ink NBSP's advance (phew!). an ink-extent width would stop the underline short of the last glyph's side bearing and diverge from browsers. I added decoration_rect_spans_the_run_advance_box test to pins this aspect.

src/lib.rs
matthargett: this actually made me go deeper in the testing with RTL, and I found an issue that I've fixed in a separate focused PR: https://github.com/femtovg/femtovg/pull/316

the combinatorics are both fun and exhausting at the same time :P

matthargett:
Closing this because it landed on master in its entirety via the [#293](https://github.com/femtovg/femtovg/pull/293) rebase-merge (ab33d9f…4c4868a plus the review-fix commits 4c4868a, 8c9ef9a, f066ff1, f222e90, 9e2a3a6). All review comments here were addressed in those commits; the text-decoration feature will ship in the next release.

PR #293, opened 2026-07-10
Expose subscript, superscript and typographic font metrics

matthargett:
Completes the `FontMetrics` TODO (`underline, strikeout, subscript, superscript metrics`) begun in #291: exposes the remaining OpenType typographic metrics so text-layout layers (parley, cosmic-text, applications) can implement `vertical-align: super/sub`, CSS `ex`/`cap` units, small-caps synthesis, and line-box layout. **Metrics only — no rendering changes**; super/subscript *rendering* is typesetting (baseline shift + size) and stays the layout layer's job.

**Stacked on #291** (text decoration) — rebased onto current master; the diff shows #291's commits beneath the two typographic-metrics commits at the top, so merge #291 first. Metrics-only, independent of the decoration rendering.

### API (all on `FontMetrics`, scaled to the requested font size like the existing accessors)

| Accessor | OpenType source |
| --- | --- |
| `subscript_size()` / `superscript_size()` → `(f32, f32)` | OS/2 `ySub/SuperscriptXSize/YSize` (MVAR-adjusted for variable fonts) |
| `subscript_offset()` / `superscript_offset()` → `(f32, f32)` | OS/2 `ySub/SuperscriptX/YOffset` |
| `x_height()` | OS/2 `sxHeight` |
| `cap_height()` | OS/2 `sCapHeight` |
| `line_gap()` | hhea (OS/2 typo gap under `USE_TYPO_METRICS`) |

PS: lolz at seeing OS/2 as a sequence of characters post-1997 XD

Script offsets are normalized to the canvas **+y-down** convention (subscript offset typically positive/down, superscript negative/up); each doc comment states the convention and the contrast with `underline_position()`'s post-table +y-up convention established in #291.

Fonts lacking the tables fall back to documented conventional ratios (script size 0.65 em, subscript drop 0.14 em, superscript rise 0.48 em, x-height 0.5 em, cap-height 0.7 em, line gap 0) shared by both metric backends. The ratios are empirically grounded: RobotoFlex's real OS/2 values are these same conventions (e.g. `ySubscriptYOffset` 287/2048 = 0.14). The swash backend exposes x-height/cap-height/leading natively and takes the shared fallbacks for the OS/2 script metrics it does not surface, so both backends agree on availability.

### Footprint

Zero rendering-path changes; zero heap allocations — `FontMetrics` remains a plain `Copy` stack struct (36 → 80 bytes, still under two cache lines on iPhone XS and Meta Quest 2), populated during the `Face::parse` table read that already runs; no new dependencies; nothing added to any per-frame path.

### Comparison vs Chromium Canary

femtovg exposes no script *rendering* (that is the layout layer's job), so these panels typeset `E=mc²` / `H₂O` **from the metrics this PR exposes**, then compose them with decorations and drop shadow. Chrome renders the same content driven with the **identical OS/2 numbers** (SVG `<text>` baseline + `baseline-shift`, `feDropShadow`), so a match confirms the metrics are correct and consumable end to end:

typographic metrics vs Chromium

Rows: superscript, subscript, then both combined with underline / strikethrough / overline, and finally with a sharp and a blurred drop shadow. Per-row difference from Chrome (max-channel > 40, the antialiasing floor):

| Row | Δ area |
| --- | --- |
| superscript | 1.2% |
| subscript | 0.5% |
| + underline | 2.1% |
| + strikethrough | 2.0% |
| + overline | 2.1% |
| + underline + sharp shadow | 3.3% |
| + underline + blur shadow | 2.1% |

Every row sits at the plain-glyph-edge antialiasing floor — script placement, decoration composition, and shadow all track Chrome with no structural difference. (The shadow rows are within the same floor because the composite's runs are spaced apart; femtovg's per-operation shadow model only diverges from a browser's whole-element `filter` when translucent shadows of *overlapping* operations superimpose, which is the offscreen-compositing/layers item on the roadmap, not a metrics concern.)

**Intentional difference — OS/2 metrics vs browser UA defaults.** These panels drive Chrome with explicit OS/2-derived sizes/offsets. Chrome's *native* `<sub>`/`<sup>` instead use its own heuristic (`font-size: smaller` + `vertical-align: super/sub`), which sizes and positions scripts differently. Exposing the font's own recommendation is the point: a layout layer can honor the type designer's metrics rather than a fixed heuristic.

### Tests

- Stripped-font fixture (minimal head/hhea/maxp TTF built in-test) asserts every documented fallback exactly — and adds coverage for #291's underline/strikeout fallbacks.
- Real-font plausibility: 0 < x-height < cap-height ≤ ascender; script sizes within (0, em); subscript points down, superscript up.
- Linear scaling: metrics at size 20 exactly half of size 40.
- Placement envelope: a superscript typeset at its recommended size lifts its cap above the base x-height while staying within the ascent, and a subscript drops within the descent depth — a magnitude check (not just sign) that guards the superscript sign-normalization against a swap or mis-scale regression.
- Suites green under the ttf-parser backend (`wgpu textlayout`: 63 passed) and the swash backend (`--no-default-features --features swash` lib: 19 passed).

### Validation

- `cargo fmt --check`; rustdoc clean
- Builds: default, `wgpu`, `--no-default-features`, `--all-features`, swash
- `cargo test --features "wgpu textlayout"`

### DIFF
--- a/src/text/font.rs
+++ b/src/text/font.rs
@@ -65,7 +65,7 @@ pub enum GlyphRendering<'a> {
 #[derive(Copy, Clone, Default, Debug)]
 struct FontFlags(u8);
 
-// TODO: underline, strikeout, subscript, superscript metrics
+// TODO: subscript, superscript metrics
 impl FontFlags {
     fn new(regular: bool, italic: bool, bold: bool, oblique: bool, variable: bool) -> Self {
         let mut flags = 0;
@@ -119,6 +119,10 @@ pub struct FontMetrics {
     ascender: f32,
     descender: f32,
     height: f32,
+    underline_position: f32,
+    underline_thickness: f32,
+    strikeout_position: f32,
+    strikeout_thickness: f32,
     flags: FontFlags,
     weight: u16,
     width: u16,
@@ -129,6 +133,10 @@ impl FontMetrics {
         self.ascender *= scale;
         self.descender *= scale;
         self.height *= scale;
+        self.underline_position *= scale;
+        self.underline_thickness *= scale;
+        self.strikeout_position *= scale;
+        self.strikeout_thickness *= scale;
     }
 
     /// Returns the distance from the baseline to the top of the highest glyph.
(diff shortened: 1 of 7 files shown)


src/text/font.rs
```diff
@@ -219,10 +380,49 @@ impl Font {
...
+        let superscript = ttf_font.superscript_metrics();
+        let fallback_script_size = (em * fallback::SCRIPT_SIZE, em * fallback::SCRIPT_SIZE);
+
         let metrics = FontMetrics {
-            ascender: ttf_font.ascender() as f32,
-            descender: ttf_font.descender() as f32,
+            ascender,
+            descender,
             height: ttf_font.height() as f32,
+            underline_position: underline.map_or(descender * 0.5, |m| m.position as f32),
```
tronical: I think the table could have `0` as value (so `Some(0)`) and in that case I think we should do a fallback. That's also what you're doing in the swash code (`!= 0` check).

tronical (commented):
Looks good to me - thanks! Just one nitpick inside :)

PR #294, opened 2026-07-11
Fix text position drift when zooming (scaled-atlas glyphs)

### LINKED ISSUE #286: Text position unstable with scaling with 0.25.1
We noticed this when trying to adopt 0.25.1 to Satty and had to revert. https://github.com/Satty-org/Satty/issues/527

Now I had an opportunity to try a repro with the demo example.

Image

This is directly after start.

Image

And the latter screenshot shows the example after zooming with ctrl+scroll. I'd say the position of the shapes remains the same, but the text has moved.

Relative to a background image (in Satty), it appears it is the text that moves and the other shapes remain in place.

Image
above is without zoom

Image
above is with zoom

matthargett:
Fixes #286 — text drifting out of position relative to shapes when zooming, a regression since 0.25.1.

### Symptom

Under a zoom (uniform scale + translation) the text moves while the vector shapes it annotates stay put, snapping as the zoom sweeps. Reproduced with the demo (`ctrl`+scroll) and, below, with a digit centered in a circle at zoom 1.531 — Chrome's canvas keeps it centered, 0.25.1 slides it ~17px off:

text drift vs Chrome

### Root cause

0.25.1 (uniformly-scaled crisp text via scale-baked atlas glyphs) rasterizes solid-color text from a glyph atlas baked at the zoom scale, quantized to 1/16 steps so small zoom changes don't churn the atlas. The quantized scale is correct for choosing the **bitmap resolution** — but it was also used to **position** the glyphs: positions were premultiplied by the quantized scale and the quads drawn under a translation-only transform, while vector geometry uses the true transform. So a glyph at world position `p` landed at `quantized·p + t` while everything else was at `true·p + t` — an error of `p·(quantized − true)` that grows with distance from the origin and snaps by `p/16` at each quantization boundary. Text far from the origin (large canvases, as in the linked Satty report) drifts the most.

### Fix

Keep rasterizing at the quantized scale (the atlas-cache benefit), but position with the true scale: present the pre-scaled quads under the residual scale `true/quantized` (always within one 1/16 step of 1.0) plus the true translation. The glyph then lands at `true·p + t` and its true on-screen size, sampling the quantized bitmap with a ≤~3% resample.

This is the standard raster-scale-vs-position decomposition mature GPU text stacks use — Skia factors the glyph matrix into `Remaining × scale` (`SkMatrix::decomposeScale`) and reconstructs true placement from the full-precision position matrix with the residual applied at vertex time; WebRender quantizes `RasterSpace::Local(scale)` for the bitmap while positioning at the true picture transform; Chrome separates raster scale from the transform tree. The original code's mistake was feeding the quantized raster scale back into positioning — exactly the coupling those designs avoid.

### Tests

`tests/text_scale_position_wgpu.rs` — headless GPU tests (skip without an adapter) that put the zoom pivot on the drawn feature so its true screen position is constant, then sweep the zoom across two 1/16 boundaries and assert nothing jumps:

- **horizontal drift** — a glyph's screen x stays put (fails by 22.8px before this change);
- **vertical drift** — the same on the y axis (fails by 13.7px);
- **co-location with vector geometry** — a glyph and a bar on the same world column keep a constant offset, i.e. text stays glued to the shapes it annotates (fails by 22.8px).

All three pass after the fix and the change is smooth across boundaries; full suite green (`cargo test --features "wgpu textlayout"`), `cargo fmt --check`, `cargo build --features wgpu`.

### Note on coverage

This class of defect — atlas text drifting from shapes under a fractional zoom — isn't covered by WPT (canvas tests run at zoom 1; reftest fuzzy tolerances absorb a few pixels of drift); the browsers guard it with engine-internal zoom reftests / golden images. The GPU tests above are the equivalent guard here.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1603,7 +1603,11 @@ where
         enum Rasterization {
             Path,
             Atlas,
-            ScaledAtlas { scale: f32, translation: (f32, f32) },
+            ScaledAtlas {
+                scale: f32,
+                true_scale: f32,
+                translation: (f32, f32),
+            },
         }
 
         // Classify the canvas transform. 1e-3 epsilon: tight enough to catch any
@@ -1611,10 +1615,10 @@ where
         let rasterization = match self.state().transform.as_uniform_scale_translation(1e-3) {
             // Rotation / skew / non-uniform / negative scale: outline rendering.
             None => Rasterization::Path,
-            Some((scale, tx, ty)) => {
+            Some((true_scale, tx, ty)) => {
                 // Quantize the baked scale so small animation steps don't churn the
                 // atlas; 1/16 steps (≈6%) are imperceptible at typical zoom levels.
-                let scale = geometry::quantize(scale, 1.0 / 16.0).max(1.0 / 16.0);
+                let scale = geometry::quantize(true_scale, 1.0 / 16.0).max(1.0 / 16.0);
                 if paint.text.font_size * scale > 92.0 {
                     // Cached bitmap would be too large.
                     Rasterization::Path
@@ -1624,6 +1628,7 @@ where
                 } else if matches!(paint.flavor, PaintFlavor::Color(_)) {
                     Rasterization::ScaledAtlas {
                         scale,
+                        true_scale,
                         translation: (tx, ty),
                     }
                 } else {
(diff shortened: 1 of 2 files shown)


RobertMueller2:
Thanks! I can confirm this fixes the issue in Satty that led to opening #286.

tronical (approved):

PR #297, opened 2026-07-13
Add start angle to conic gradients (createConicGradient parity)

### LINKED ISSUE #239: Add gradient dithering
Gradients between two very similar colors create harsh banding which becomes more noticeable the larger the distance the gradient covers.

Dithering is the process of creating in-between colors using an optical illusion to make sub-band transitions possible.

Image

Currently, Femtovg does not support dithering like other vector graphics libraries, leading to noticeable bands:

Image

Implementing dithering should be as simple as adding additional virtual stops between the gradient stop points with pre-set dithering patterns, see (https://en.wikipedia.org/wiki/Ordered_dithering)[https://en.wikipedia.org/wiki/Ordered_dithering]

This would greatly improve the quality of femtovg's rendering and enable more sleek and polished looking renders.

matthargett:
Adds a start angle to conic gradients, giving Canvas `createConicGradient(startAngle, x, y)` parity (and SVG 2 conic alignment). Today femtovg's conic gradient always begins at the positive x-axis; this lets the caller rotate where offset 0 sits.

### Also in this PR: gradient dithering (closes #239)

Since this PR already works in the shared gradient shader path, it also adds
screen-space **ordered dithering** to gradients, addressing the banding reported
in #239. Gradients between close colours band under 8-bit quantization; a
sub-LSB interleaved-gradient-noise offset before the framebuffer write spreads
the rounding spatially and breaks the bands. Applied to every gradient type
(linear/radial box, conic, and their multi-stop LUT variants) on both backends,
with no new uniforms or memory (a few ALU ops per fragment); solid regions are
unaffected. Covered by a wgpu test (`gradient_dither_wgpu.rs`): a close-colour
gradient has 0/220 columns dithered before, most after.

gradient dithering before/after

### API

`Paint::conic_gradient_with_angle(cx, cy, start_angle, start, end)` and `Paint::conic_gradient_stops_with_angle(cx, cy, start_angle, stops)`; the existing angle-less constructors delegate with `start_angle = 0.0` (unchanged behavior). `start_angle` is in radians, measured clockwise from +x, matching the Canvas convention. A non-finite angle (NaN/∞) is treated as `0.0` at the API edge, like the `set_shadow_*` setters.

### Comparison vs Chrome

Six-stop wheels at start angles 0°, 90°, 180° — femtovg vs Chrome's `createConicGradient(angle, cx, cy)` with the same stops. The seam/offset-0 position tracks Chrome at each angle (difference is edge antialiasing only, 0.78% of pixels above the AA floor):

conic start-angle vs Chrome

Composition testing (a conic gradient inside a rounded scissor, with text + drop shadow layered on top) surfaced a pre-existing wgpu-backend bug: the WGSL conic cases `return`ed their colour directly, *before* the scissor-mask + stroke-AA multiply every other fill type falls through to — so conic fills ignored the scissor (rounded corners included) and had no edge antialiasing. The GLSL backend was already correct. Fixed and added a pinning test which fails before the fix and passes after.

The composed scene now clips correctly and matches Chrome (`createConicGradient` + `roundRect` clip, same numbers) at 0.71%:

conic clipped by rounded scissor, text+shadow on top

GLSL, WGSL and the Rust `UniformArray`/`Params` were verified byte-consistent, and a test (`rounded_scissor_and_conic_start_angle_coexist_in_params`) asserts an angled conic and a rounded scissor set their respective slots without messing eachother up. The GPU phase test renders through the current shader (which now also contains the shadow-blur uniforms), so it validates the byte offset end to end, not just the glue between the pieces.

very fun to come up with compositions of these features and stay within the memory/compute limit goals :D

### Back-compat

`start_angle` is `#[serde(default)]`, so paints serialized before this field deserialize to a `0.0` start angle (previous behavior). Covered by a round-trip test.

### Tests

- `conic_gradient_start_angle_is_plumbed_into_params` — the angle reaches `Params`/the uniform slot.
- `rounded_scissor_and_conic_start_angle_coexist_in_params` — no slot collision with #280.
- `conic_gradient_start_angle_matches_canvas_convention` — GPU render: offset 0 at +x for angle 0, advancing clockwise (mirrors WebKit's `fast/canvas/canvas-conic-gradient-angle`).
- `conic_gradient_non_finite_start_angle_is_treated_as_zero` — the API-edge guard.
- serde back-compat round-trip.

### Validation

- `cargo fmt --check`
- `cargo test --features "wgpu textlayout"` (both backends green)

Filed as a draft pending review bandwidth; no dependency on other open PRs.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,6 +59,7 @@ textlayout = ["dep:rustybuzz", "dep:unicode-bidi", "dep:unicode-segmentation", "
 swash = ["dep:swash"]
 
 [dev-dependencies]
+serde_json = "1.0"
 winit = { version = "0.30.12" }
 euclid = "0.22.3"
 rand = "0.10"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2714,6 +2714,35 @@ fn conic_gradient_start_angle_is_plumbed_into_params() {
     assert_eq!(params.conic_start_angle, angle);
 }
 
+/// The scissor radius (frag[12].w) and the conic start angle (frag[13].x) live
+/// in adjacent uniform slots. A draw that uses both a rounded scissor and a
+/// conic gradient with a start angle must carry each value independently in the
+/// same `Params`, so neither feature can clobber the other's slot.
+#[test]
+fn rounded_scissor_and_conic_start_angle_coexist_in_params() {
+    use renderer::ShaderType;
+
+    let renderer = RecordingRenderer::default();
+    let recorded_commands = renderer.last_commands.clone();
+    let mut canvas = Canvas::new(renderer).unwrap();
+    canvas.set_size(100, 100, 1.0);
+
+    canvas.rounded_scissor(10.0, 10.0, 80.0, 80.0, 12.0);
+
+    let angle = std::f32::consts::FRAC_PI_2;
+    let paint = Paint::conic_gradient_with_angle(50.0, 50.0, angle, Color::rgb(255, 0, 0), Color::rgb(0, 0, 255));
+    let mut path = Path::new();
+    path.rect(0.0, 0.0, 100.0, 100.0);
+    canvas.fill_path(&path, &paint);
+    canvas.flush_to_output(());
+
+    let commands = recorded_commands.borrow();
+    let params = first_draw_params(&commands);
+    assert_eq!(params.shader_type, ShaderType::FillGradientConic);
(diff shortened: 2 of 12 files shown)


tronical:
This looks interested. Don't hesitate to mark it as ready for review (!= draft) when it's ready :)

tronical (approved):
Thanks :)

Issue #299, opened 2026-07-18
WGPU renderer panics uploading an HtmlImageElement whose attributes exceed its natural size


yebei199:
## Summary

`ImageSource::HtmlImageElement` means a different size to each backend, and the WGPU renderer turns the disagreement into a panic. Uploading an SVG-backed `<img>` whose width/height attributes have been enlarged for hidpi output aborts the process:

```
panicked at wgpu-29.0.4/src/backend/webgpu.rs:2697:14:
called `Result::unwrap()` on an `Err` value: JsValue(OperationError:
Failed to execute 'copyExternalImageToTexture' on 'GPUQueue':
Copy rect is out of bounds of external image
```

## Where the disagreement comes from

`ImageSource::dimensions()` reads `element.width()` for an `HtmlImageElement`, so it returns the attribute size. The OpenGL renderer uploads through `tex_sub_image_2d_with_html_image`, and the browser rasterizes an SVG at those attributes, so the attribute size is the truth there. The WGPU renderer passes the same number to `copy_external_image_to_texture` as the copy extent, but WebGPU takes the source size of an `HTMLImageElement` from its natural size and validates the rect against that. Enlarging the attributes therefore produces a rect larger than the source.

Setting the attributes is how a caller asks for a sharper rasterization of a scalable source, and it works under WebGL. Slint does exactly this for hidpi displays, which is how I ran into it.

## Evidence that the attributes buy nothing on WebGPU

This page uses a 16x16 SVG with its attributes set to 20x20, matching a 1.25 scale factor:

```html
<script type="module">
const SVG = `<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16" viewBox="0 0 16 16">
  <rect width="16" height="16" fill="black"/><path d="M0 16 L16 0" stroke="white" stroke-width="1"/></svg>`;
const img = new Image();
img.src = URL.createObjectURL(new Blob([SVG], { type: "image/svg+xml" }));
await img.decode();
img.width = 20; img.height = 20;
console.log(img.naturalWidth);                // 16

const device = await (await navigator.gpu.requestAdapter()).requestDevice();
const tex = (n) => device.createTexture({ size: [n, n], format: "rgba8unorm",
  usage: GPUTextureUsage.COPY_DST | GPUTextureUsage.COPY_SRC | GPUTextureUsage.RENDER_ATTACHMENT });

device.queue.copyExternalImageToTexture({ source: img }, { texture: tex(20) }, [20, 20]);
[...]

PR #301, opened 2026-07-19
Rasterize resized HtmlImageElement uploads through a canvas

yebei199:
Fixes the panic tracked in #299: WebGPU copies an `HtmlImageElement` from its
natural size, so a caller that enlarges the `width`/`height` attributes asks
for a copy extent larger than the source and validation rejects it. Slint does
exactly this to get sharper icons on hidpi displays.

When the attribute size differs from the natural size, the upload now goes
through an offscreen canvas drawn at the attribute size, reusing the
`HTMLCanvasElement` path added in #300. Matching sizes stay on the direct path,
and the OpenGL backend is untouched, since the attribute size is already the
truth there. Callers need no change.

The condition triggers on any mismatch, so an element whose attributes are
*smaller* than its natural size now takes the canvas route as well. My reading
of `copyExternalImageToTexture` is that the direct path crops from the source
origin there rather than scaling down, which would make this a fix too, but I
have only measured the enlarging case. If you would rather keep the change
strictly to the reported panic, narrowing the condition to `>` is a one line
edit and I will push it.

## On ImageBitmap

@tronical suggested an intermediate `ImageBitmap` in #299, and this does the
same thing with a different primitive. `createImageBitmap` returns a Promise,
while `Renderer::update_image` is a synchronous trait method. A wasm main
thread cannot wait for that Promise to resolve: the event loop that would
settle it is the one being blocked. Making the path async would spread up
through `Canvas::update_image`, into the glyph atlas uploads in `text.rs`, and
therefore into `fill_text`, forcing all three backends to change for a branch
that only ever runs on wasm with WebGPU.

`drawImage` is synchronous and produces the same result. It also has evidence
behind it: the measurement in #299 reads back three distinct shades along the
diagonal of a 20x20 upload, so the SVG really is re-rasterized rather than
stretched. `resizeWidth`/`resizeHeight` on `createImageBitmap` is specified as
resizing the decoded bitmap, which does not guarantee re-rasterization.

## Testing

Verified by hand in a browser, since the path needs wasm, WebGPU and a DOM
`` at once. The repository has no wasm test harness to hook into, and the
existing headless GPU tests run against a native adapter where
`HtmlImageElement` does not exist. Happy to add one if you want the tooling.

Two edge cases are left alone: an image that has not finished loading, and an
SVG with no intrinsic size, both give `drawImage` nothing to draw and upload a
blank canvas where the direct path would raise an error. Guarding those changes
the error semantics of the API, so it seemed better as a separate discussion.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ glutin = { version = "0.32", optional = true, default-features = false }
 web_sys = { version = "0.3", package = "web-sys", features = [
     "WebGlContextAttributes",
     "HtmlImageElement",
+    "CanvasRenderingContext2d",
     "WebGl2RenderingContext",
 ] }
 wasm-bindgen = "0.2"
--- a/src/renderer/wgpu.rs
+++ b/src/renderer/wgpu.rs
@@ -239,6 +239,35 @@ pub struct WGPURenderer {
     pipeline_cache: Rc<RefCell<HashMap<PipelineState, CachedPipeline>>>,
 }
 
+/// Rasterizes an image element into an offscreen canvas at the given size.
+#[cfg(target_arch = "wasm32")]
+fn rasterize_to_canvas(
+    element: &web_sys::HtmlImageElement,
+    size: crate::image::Size,
+) -> Result<web_sys::HtmlCanvasElement, crate::ErrorKind> {
+    use wasm_bindgen::JsCast;
+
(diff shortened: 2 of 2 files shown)


tronical (approved):
Thanks, good point about the promise. This is heavy, but it's only done once per update_image.

PR #302, opened 2026-07-19
perf(wgpu): one frame uniform buffer with dynamic offsets, and fewer redundant calls

yebei199:
Addresses #305. This supersedes #303, which is now closed: its commit is the
first one here, so there is nothing left to review separately.

## What the per-draw cost is

A draw on this backend created six GPU objects. Two samplers and a texture view
never varied, and the other three came from the uniforms.

`create_buffer_init` for a draw's 224 bytes of uniforms is alloc, map, copy and
unmap, four trips into the GPU process. The bind group missed its cache on
nearly every draw because the cache key contained those uniforms, so two
rectangles differing only in position rebuilt it.

## What this does

The first commit caches the samplers and the empty texture view, which removes
four of the six objects and buys about 10%.

The rest needs the binding layout to change. Binding one frame-sized uniform
buffer with dynamic offsets, sized from an upper bound on the draw count before
recording starts, leaves the bind group's key describing only the texture
bindings, so consecutive draws against the same image reuse it. 1000 rectangles
go from 2.0fps to 108.4fps.

Then uniforms upload once per command rather than once per draw, which is what
`opengl.rs` already does by calling `set_uniforms` outside the drawable loop,
and the frame's slots are staged in a `Vec` and uploaded in one call.

Then the vertex buffer stays resident instead of being rebuilt every frame, and
`set_pipeline`, `set_stencil_reference` and `set_bind_group` are skipped when
they would be no-ops, the way `select_main_program` already guards program
rebinds.

The last commits are cleanup: the uniform buffer is sized per command rather
than per drawable, since the drawables inside a command reuse its params, which
cuts a 2000-rectangle frame's reservation from 8 MiB to 1 MiB and comes with a
test that fails if the bound stops covering a frame.

Skipping those state calls is only sound within one render pass, since a new
pass starts with default state. `RenderPassBuilder` counts passes and the caches
invalidate when the count moves. Without that guard the saving draws wrong, and
only on scenes that render into layers.

## Measurements

Method is described in #305. Reproduce with `examples/many_rects.rs` from #304.

| rectangles | before | after | GL renderer |
| --- | --- | --- | --- |
| 0 | 4.25ms | 2.11ms | 0.28ms |
| 100 | | 2.45ms | 0.35ms |
| 200 | 103.06ms, 8.1fps | | |
| 300 | | 2.12ms | 0.41ms |
| 600 | | 3.01ms, 144fps | 0.71ms, 144fps |
| 1000 | 434.46ms, 2.0fps | | |

Per element the slope goes from roughly 500us to about 1.2us, against 0.72us
for the GL renderer. What remains is the idle cost: 2.11ms a frame with one
rectangle on screen against GL's 0.28ms, of which `vkQueueSubmit` is 0.06ms, so
it is still Dawn's CPU side.

## Status

Measured only on wasm through Chrome. I have not measured native, where
the same allocations are cheaper and the change may show little.

Rendering was checked against screenshots of a real interface with layers,
blurred shadows, text and SVG paths, not only against the frame numbers. A
wrong dynamic offset or a skipped state call draws garbage at a perfectly good
frame rate.


matthargett:
I just did a CPU profiling and optimization pass, but haven't staged the PR yet. This work is totally complementary, so no conflicts and this generally looks good for my targets (Apple A10X efficiency CPU cores and GPU),and  the contiguous per-frame uniform staging is also the shape a future native multi_draw_indirect path would want; the only swap is dynamic-offset → storage-array indexing, so this all LGTM! :D

yebei199:
Good to know it holds up on A10X. That is the case I could not measure.

The storage-array swap cannot land here, though. The examples request
`Limits::downlevel_webgl2_defaults()`, which pins
`max_storage_buffers_per_shader_stage` to 0, so a storage binding fails
validation on the WebGL2 backend outright. The dynamic offset is also the only
channel carrying the per-draw slot index right now: draws are non-instanced
`draw(v, 0..1)`, so `instance_index` is always 0. Feeding that index another way
costs more per draw than the offset does, until `multi_draw_indirect` makes
`first_instance` free.

So it wants to sit behind `Features::MULTI_DRAW_INDIRECT`, with the uniform path
kept as the WebGL2 fallback. Happy to pick that up once your profiling work
lands, or to leave it to you if you would rather own it.

src/renderer/wgpu.rs
tronical: Is it possible that this is unused now?

tronical (commented):
Very nice work. Thanks! One little question inside :)

src/renderer/wgpu.rs
```diff
@@ -1722,27 +1846,56 @@ impl CommandToPipelineAndBindGroupMapper {
         image: Option<ImageOrTexture>,
         glyph_texture: GlyphTexture,
     ) {
+        // A new render pass resets state, so nothing set on the previous one still counts.
+        if self.current_pass_generation != Some(render_pass_builder.pass_generation) {
```
tronical: Instead of this current_pass_generation, wouldn't it be easier to move the fields here (current_pipeline_state, current_stencil_reference, current_bounding_offset) straight into RenderPassBuilder?

src/renderer/wgpu.rs
```diff
@@ -1722,27 +1846,56 @@ impl CommandToPipelineAndBindGroupMapper {
...
+        let uniforms = UniformArray::from(params);
+        if self.current_uniforms.as_ref() != Some(&uniforms) {
+            let offset = self.uniform_staging.len() as u64;
+            self.uniform_staging
+                .extend_from_slice(bytemuck::cast_slice(uniforms.as_slice()));
+            self.uniform_staging.resize((offset + self.uniform_stride) as usize, 0);
+            self.current_uniforms = Some(uniforms);
+            self.current_uniform_offset = offset;
+        }
+        let offset = self.current_uniform_offset as u32;
```
tronical: Perhaps you could remove `self.current_uniform_offset` and instead compute it, based on `self.uniform_staging.len()` minus the stride.

src/renderer/wgpu.rs
tronical: Perhaps you could move this cache lookup into the if, i.e. look up only if the pipeline changes.

tronical: Can this still happen after your change?

src/renderer/wgpu.rs
```diff
@@ -229,7 +258,11 @@ pub struct WGPURenderer {
 
     screen_view: [f32; 2],
 
-    empty_texture: wgpu::Texture,
+    empty_texture_view: wgpu::TextureView,
+    sampler_cache: Rc<RefCell<HashMap<crate::ImageFlags, wgpu::Sampler>>>,
```
tronical: This appears a couple of types - a good candidate for a type alias :)

src/renderer/wgpu.rs
yebei199: Yes. The tracking moved to the mapper and the builder's copy was only ever written. Removed in 606bbab.

src/renderer/wgpu.rs
yebei199: Done in 00c250d. recreate_render_pass resets the three fields directly and the counter is gone. begin_render_pass has exactly one call site, so no pass can start without the reset.

src/renderer/wgpu.rs
yebei199: Done in b4e07de. The slot bound this draw is always the last one staged, so the offset falls out of uniform_staging.len().

src/renderer/wgpu.rs
yebei199: Done in 47f37b3. The retain sweep still sees every pipeline used in a frame: each pass starts with no bound pipeline, so the first draw through each state performs the lookup that marks it accessed.

yebei199: No. The buffer is created at MIN_VERTEX_BYTES and only ever grows. Guard removed in c896d9f.

src/renderer/wgpu.rs
yebei199: Added SamplerCache in 2717ec5.

tronical (approved):

PR #303, opened 2026-07-19
perf(wgpu): cache the samplers and empty texture view instead of rebuilding them per draw

yebei199:
Addresses part of #305, with #302 stacked on it for the rest.

## What this does

On a bind group cache miss the backend creates a uniform buffer, two samplers,
two texture views and a bind group. The two samplers and two views are of the
1x1 empty texture and never vary with the scene, so this keeps the view on the
renderer and caches samplers by the three flags that reach the descriptor.

That removes four of the six objects created per draw. It buys about 10% on its
own, which is what showed that the uniform buffer and the bind group carry the
rest. Those need the binding layout to change and are #302.

## Measurements

Method and full figures are in #305. Reproduce with `examples/many_rects.rs`
from #304.

| rectangles | before | with this PR |
| --- | --- | --- |
| 200 | 103.06ms, 8.1fps | 97.12ms, 9.9fps |
| 1000 | 434.46ms, 2.0fps | 378.47ms, 2.3fps |

Per-rectangle sampler and texture view creation goes from two of each to zero.

## Status

Measured only on wasm through Chrome, where crossing the JS boundary makes each
resource creation expensive. Native was not measured and may show little.

### DIFF
--- a/src/renderer/wgpu.rs
+++ b/src/renderer/wgpu.rs
@@ -213,6 +213,24 @@ pub struct Image {
     info: ImageInfo,
 }
 
+/// The only flags a sampler descriptor reads. Keying on all of `ImageFlags` would miss on the rest.
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq)]
+struct SamplerKey {
+    repeat_x: bool,
+    repeat_y: bool,
+    nearest: bool,
+}
+
+impl From<crate::ImageFlags> for SamplerKey {
+    fn from(flags: crate::ImageFlags) -> Self {
+        Self {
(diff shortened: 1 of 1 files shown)


yebei199:
Closing this one: its commit is the first of #302, so having both open means the same change gets reviewed twice. Nothing is lost, and #302 now describes the sampler and view caching as its own first step.

The follow-up cleanup that landed here today is in #302 as well: the sampler cache keys on the three `ImageFlags` bits a sampler descriptor actually reads instead of a struct mirroring them, and the `empty_texture` field the view replaced is gone.

PR #306, opened 2026-07-27
Report out-of-range image updates instead of aborting

matthargett:
## Summary

`Renderer::update_image` is specified to reject a copy it cannot perform. The `Void`
renderer, the OpenGL backend and the in-tree test recorder each check that the copy fits
inside the destination image; OpenGL additionally checks that the source's pixel format
matches the one the image was created with. They return `ErrorKind::ImageUpdateOutOfBounds`
and `ErrorKind::ImageUpdateWithDifferentFormat`.

The WGPU backend performed **neither** check. It passed the copy straight to
`Queue::write_texture`, or on wasm to `copyExternalImageToTexture`, both of which fail
graphics API validation — and wgpu surfaces that as a panic. So a recoverable caller
mistake takes down the process instead of the single call. On wasm it takes down the
application: the instance is dead, the canvas keeps its last painted frame, and no further
input is processed.

## Reproducing it

This is not wasm-specific — it reproduces natively on any adapter. A 32x32 source written
at (8, 8) into a 32x32 image, which every other backend reports as
`ImageUpdateOutOfBounds`:

```
thread 'update_image_past_the_edge_is_rejected' panicked at wgpu-30.0.0/src/backend/wgpu_core.rs:2068:22:
wgpu error: Validation Error

Caused by:
  In Queue::write_texture
    Copy of X 8..40 would end up overrunning the bounds of the Destination texture of X size 32
```

And a `Gray8` source into an `Rgba8` image, which OpenGL reports as
`ImageUpdateWithDifferentFormat`:

```
  In Queue::write_texture
    Number of bytes per row is less than the number of bytes in a complete row
```

Both tests in this PR fail that way on `master` and pass with the change.

## The fix

Three implementations had each derived the same rule and the fourth had none, which is how
the divergence arose in the first place. Rather than write it a fourth time, this moves it
into `ImageSource::check_update()` and calls that from all of them, so a backend cannot
quietly omit it. `Renderer::update_image` now documents that implementations should call
it. The helper is public because `Renderer` is public and out-of-tree backends need the
same guarantee.

Behaviour per backend:

| | before | after |
| --- | --- | --- |
| OpenGL | bounds + format | unchanged |
| `Void` | bounds | bounds + format |
| WGPU | **neither** | bounds + format |

The check also saturates when adding the origin to the source size, so an origin near
`usize::MAX` is reported rather than wrapping into a range that looks valid — in a release
build the previous `x + size.width` wrapped and let an enormous origin through to the
graphics API.

## Relationship to #300 / 1c174fd

This is complementary to the `HtmlCanvasElement` source and the resized-`HtmlImageElement`
rasterization that went into 0.26.0. Those fixed the **source** side of one specific case:
an `HtmlImageElement` enlarged for hidpi has a copy extent larger than the natural size
WebGPU reads, so the copy rect fell outside the *source*. This fixes the **destination**
side, which is unguarded for every caller and every geometry, and which no rasterization
step can catch. The two do not overlap: the wasm paths keep rasterizing exactly as they do
today, and now also validate against the destination first.

Reported downstream as slint-ui/slint#12540, where the hidpi SVG case aborted a Slint
application in the browser. Its author suggested exactly this, independently of the narrow
fix: *"femtovg's WGPU renderer could clamp the copy extent ... so that other consumers of
the same API do not hit a panic."* Returning the documented error rather than clamping
keeps it consistent with the other backends, and does not silently paint something the
caller did not ask for.

## Tests

- `tests/image_update_contract.rs` — the contract over the `Void` renderer, so it holds on
  every platform and runs in CI without a GPU: past the edge, a source larger than the
  destination, an overflowing origin, a mismatched format, and the copies that do fit
  (including one flush with the far corner) still succeeding.
- `tests/image_update_bounds_wgpu.rs` — the same cases against the WGPU backend, which is
  where they aborted. Skips when no adapter is available.

## Verification

`cargo fmt --all -- --check`, `cargo clippy --all-features --no-deps` (warning count
unchanged from `master`), `cargo doc --no-deps --all-features`, default and `--all-features`
test suites (71 and 97 passing), and builds for default / `--all-features` /
`--no-default-features` / `wasm32-unknown-unknown` / `wasm32-unknown-unknown --features wgpu`,
the last being the configuration in the downstream report.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ All notable changes to this project will be documented in this file.
   the corner square instead of bulging it outwards or emitting NaN coordinates,
   and an infinite radius rounds as far as the box allows. This changes rendering
   for shapes whose radii did not fit.
+- Fixed the WGPU renderer aborting instead of reporting an error when
+  `update_image()` is given a copy that reaches past the destination image, or a
+  source in a different pixel format. Both now return
+  `ErrorKind::ImageUpdateOutOfBounds` and
+  `ErrorKind::ImageUpdateWithDifferentFormat`, as the OpenGL and `Void`
+  renderers already did. The shared check is exposed as
+  `ImageSource::check_update()` for out-of-tree renderers, and reports rather
+  than overflows for an origin close to `usize::MAX`.
 
 ## [0.26.0] - 2026-07-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+- Fixed the WGPU renderer aborting instead of reporting an error when
+  `update_image()` is given a copy that reaches past the destination image, or a
+  source in a different pixel format. Both now return
+  `ErrorKind::ImageUpdateOutOfBounds` and
+  `ErrorKind::ImageUpdateWithDifferentFormat`, as the OpenGL and `Void`
+  renderers already did. The shared check is exposed as
+  `ImageSource::check_update()` for out-of-tree renderers, and reports rather
+  than overflows for an origin close to `usize::MAX`.
+
 ## [0.26.0] - 2026-07-20
 
 - Added Canvas 2D drop shadows for fills, strokes and text. New `Canvas` methods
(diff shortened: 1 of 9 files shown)


tronical (approved):

tronical:
Long words for "added missing check to wgpu renderer" ;-). Please rebase though to resolve the conflict.

matthargett:
> Long words for "added missing check to wgpu renderer" ;-). Please rebase though to resolve the conflict.

I try to make the headline semantically meaningful since the check addition can be seen in the diff. if someone on an older version runs into the abort and searchea for "abort" in issues/PRs, they can (hopefully) find the merged fix and discussion

PR #307, opened 2026-07-27
Scale rounded rect radii together instead of per axis

matthargett:
## Summary

I was chasing some pixel differences in BabylonNative's screenshot suite, and found this problem that I previously thought was aliasing edge noise in the screenshot diff.

`rounded_rect_varying()` fitted an oversized radius by clamping each axis on its own:

```rust
let rx_tl = rad_top_left.min(halfw) * w.signum();
let ry_tl = rad_top_left.min(halfh) * h.signum();
```

That keeps the corner inside the box, but it turns a circular corner into an **ellipse**. On a
300×75 rectangle with a radius of 75 it produces `rx = 75`, `ry = 37.5`, so the ends come out
stretched instead of semicircular:

before, after and Chrome

CSS Backgrounds and Borders Level 3 ([§4.5 Overlapping curves](https://www.w3.org/TR/css-backgrounds-3/#corner-overlap))
and the Canvas [`roundRect()`](https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-roundrect)
algorithm specify the same reduction: take

```
f = min(side length / sum of that side's two radii)   over the four sides
```

and, if `f < 1`, multiply **every** radius by it. One factor for all of them is what keeps each
corner the shape it was asked for — the CSS text says so directly: *"This formula ensures that
quarter circles remain quarter circles."* Blink implements exactly this, as a single `factor`
applied to both axes of all four corners, in `FloatRoundedRect::ConstrainRadii`
(`third_party/blink/renderer/platform/geometry/float_rounded_rect.cc`), and shares that code
between CSS `border-radius` and canvas `roundRect`.

## What else the specified rule fixes

- **A corner may take a whole side when the corner beside it is square.** The sum along a side
  is what is bounded, not each radius, so a radius of 100 on a 100×100 box is legal. Clamping to
  half the dimension cut it to 50. This is WPT `border-radius-sum-of-radii-001` case 3.
- **Radii are reduced in proportion to one another**, including ones that would have fitted on
  their own — the spec calls this out too. Case 4 of that test reduces the bottom corners from 30
  to 25 purely because the top side overflows.
- **Negative and NaN radii.** Previously a negative radius reached the geometry and bulged the
  corner outwards, and a NaN compared false against the small-radius test, then survived `f32::min`
  (which returns the non-NaN operand) and spread through every coordinate of the shape. Both now
  leave the corner square. Canvas throws a `RangeError` for a finite negative radius; this API has
  no way to report that, so it treats it as no rounding rather than emitting geometry that cannot
  be drawn. An infinite radius is treated as the largest finite one and reduced to fit, since
  asking for more rounding than there is room for is the usual way to ask for a fully round shape.

## Tests

`tests/rounded_rect_radii.rs`, all at the geometry level so they need no GPU and run everywhere.
The radii are read back from the emitted path's on-curve endpoints, so they are exact rather than
sampled.

Ported from the web platform tests:

- **`wpt_css_border_radius_sum_of_radii`** — all eight cases of
  `css/css-backgrounds/border-radius-sum-of-radii-001`, whose reference values are the used radii
  the rule must produce.
- **`wpt_a_factor_of_exactly_one_reduces_nothing`** — the boundary from `-002`: radii that exactly
  fill a side give `f == 1` and nothing is reduced, since the rule applies only below one.
- **`wpt_canvas_intersecting_radii_reduce_to_a_circle`** — the geometry behind
  `2d.path.roundrect.radius.intersecting.1` and `.2`.
- **`wpt_negative_extents_round_the_origin_corner`** — a pin, not a fix: `2d.path.roundrect.negative`
  expects the top left radius to round the corner at the origin for a negative width or height.
  It passes before and after, and is here because the reduction must run on side lengths; using the
  signed extents would divide a negative width by a side summing to zero and scale every radius by
  minus infinity.

Ported from Gecko's reftests, both of which a per-axis clamp fails:

- **`gecko_corner_3_keeps_full_width_radii`** — `layout/reftests/border-radius/corner-3.html`: a
  30×130 box whose two right corners are rounded by 30, the full width, because each horizontal
  side carries only one of them. Clamping halves them to 15.
- **`gecko_border_reduce_height_scales_the_wide_axis_too`** — `border-reduce-height.html`, added
  alongside Gecko's implementation of this rule: a 60×20 box reduces 5, 20, 5, 20 by 0.8 to
  4, 16, 4, 16. The factor comes from the short vertical sides yet still narrows corners that the
  60 units of width had room for, which is the part no per-side or per-axis correction reproduces.

Guarding the precision the reduction newly depends on:

- **`enormous_radii_round_as_far_as_they_fit`** — radii of 1e30, `f32::MAX` and infinity on a
  100×100 box all round to 50. Adding two radii together is new work that the old clamp never did,
  and in single precision two large ones overflow to infinity, making the factor zero and squaring
  off the whole shape. Skia adds them as doubles for this reason
  (`SkRRect::scaleRadii`, crbug.com/463920), as does Gecko in `nsIFrame::ComputeBorderRadii`.
  Flutter, which scales in single precision, has the visible form of it:
  `BorderRadius.circular(double.infinity)` gives a square box (flutter/flutter#124525).

One note on the canvas pair. Both of those WPT tests sample pixels well inside the shape, and I
checked every asserted pixel against both behaviours: an implementation that clamps each axis
separately draws a 40×25 and a 50×25 ellipse and **still passes both tests**. Copying their pixel
assertions would add tests that cannot fail on this bug, so that case asserts the radii instead.

Plus the behaviour these were written for: a scalar radius never producing an ellipse across a
range of box shapes, radii that fit being left alone, the factor coming from the worst side rather
than the first that overflows, unusable radii leaving a square corner without leaking a non-finite
coordinate, and a degenerate box staying finite.

Nine of the sixteen fail on `master`; the rest pin behaviour that must not change.

## Verification

Expected values were cross-checked two ways: against Chrome's `roundRect()` (rendered, sampling
the leftmost covered pixel per row — the fixed output matches to within a pixel of bezier and
antialiasing noise), and against an independent implementation of the spec formula run over eight
geometries, which agrees exactly on all four radii in every case.

`cargo fmt --all -- --check`, `cargo clippy --all-features --no-deps` (warning count unchanged from
`master`), default and `--all-features` test suites (82 and 104 passing), and builds for default /
`--all-features` / `--no-default-features` / `wasm32-unknown-unknown`.

## Compatibility

This changes rendering for shapes whose radii did not fit; shapes whose radii already fitted are
untouched. That is the point of the change, but it is a visible difference and worth a note in the
release.

Reported downstream as [slint-ui/slint#6506](https://github.com/slint-ui/slint/issues/6506), where a
Slint `Rectangle` with `border-radius: self.height` renders squashed. Slint passes the radius
straight through to `rounded_rect()` without clamping, so the shape comes entirely from here. It has
also been quietly costing pixel differences in BabylonNative's screenshot tests, where the corner
error reads as antialiasing noise rather than a wrong shape.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,10 @@ All notable changes to this project will be documented in this file.
   the Canvas `roundRect()` algorithm both specify, so corners keep their shape:
   a radius larger than half the height now gives a fully rounded end rather than
   a squashed one. A corner may also use a whole side when the corner next to it
-  is square, which the previous clamp cut in half. Negative and non-finite radii
-  leave the corner square instead of bulging it outwards or emitting NaN
-  coordinates. This changes rendering for shapes whose radii did not fit.
+  is square, which the previous clamp cut in half. Negative and NaN radii leave
+  the corner square instead of bulging it outwards or emitting NaN coordinates,
+  and an infinite radius rounds as far as the box allows. This changes rendering
+  for shapes whose radii did not fit.
 
 ## [0.26.0] - 2026-07-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,18 @@
 # Changelog
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+- Fixed `Path::rounded_rect()` and `rounded_rect_varying()` flattening corners
+  into ellipses when a radius did not fit. Radii that overlap along a side are
+  now reduced by one common factor, as CSS Backgrounds and Borders Level 3 and
+  the Canvas `roundRect()` algorithm both specify, so corners keep their shape:
+  a radius larger than half the height now gives a fully rounded end rather than
+  a squashed one. A corner may also use a whole side when the corner next to it
+  is square, which the previous clamp cut in half. Negative and non-finite radii
+  leave the corner square instead of bulging it outwards or emitting NaN
+  coordinates. This changes rendering for shapes whose radii did not fit.
+
 ## [0.26.0] - 2026-07-20
 
 - Added Canvas 2D drop shadows for fills, strokes and text. New `Canvas` methods
--- a/src/path.rs
+++ b/src/path.rs
@@ -453,8 +453,10 @@ impl Path {
     /// This is the "overlapping curves" rule shared by CSS backgrounds and
     /// borders and by the Canvas `roundRect()` algorithm.
     ///
-    /// A radius that is negative or not finite does not describe a corner and is
-    /// treated as zero, leaving that corner square.
+    /// A negative radius, or one that is NaN, does not describe a corner and is
+    /// treated as zero, leaving that corner square. An infinite radius asks for
+    /// as much rounding as there is room for and is reduced to fit, like any
+    /// other radius that is too large.
     pub fn rounded_rect_varying(
         &mut self,
         x: f32,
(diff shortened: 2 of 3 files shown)


tronical (approved):
Lovely, thanks!

PR #314, opened 2026-08-11
Added elliptical gradients

R-Cramer4:
This will allow elliptical radial gradients like in CSS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ All notable changes to this project will be documented in this file.
   drawn width changed law with the zoom, the font size, and even the paint
   flavor. All user-space text quantities now cross through the baked scale at
   one place, and a zoom-invariance test suite holds the regime seams.
+- Added elliptical radial gradients, matching CSS's `radial-gradient(ellipse ...)`.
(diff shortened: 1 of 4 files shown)


tronical (commented):
Thanks! The commit message says "Added radial gradients", but they exist already. This PR adds support for elliptical radial gradients. Please adjust.

Implementation wise, I think this can be dramatically simplified. Instead of adding new shader variants, the ellipses could be represented as a regular radial gradient and the `paint_mat` just squashing the rendering so that the circle becomes the desired ellipse.

Or am I missing something here?

R-Cramer4:
It can be moved to `paint_mat`, I didn't notice that when I was implementing it. One thing I wasn't too sure about was having the public api have so many radius parameters. There is the option of changing it to `in: (f32, f32)` instead of `in_rx: f32, in_ry: f32`.

matthargett:
heya! I had a draft for this waiting in my PR queue here: https://github.com/rebeckerspecialties/femtovg/pull/7

by all means, go for it in your own style, but I would suggest checking out the tests for the invariants I discovered when comparing screenshot rendering against firefox and chromium, and looking in their past regressions to make sure we don't back into the same conformance issues :D

tronical (approved):

PR #316, opened 2026-08-16
Report user-space metrics from Canvas::measure_font

matthargett:
`Canvas::measure_font()` multiplied its result by the canvas's internal quantized glyph-rasterization scale (`font_scale()`) and the DPI factor, while `measure_text()`, `TextContext::measure_font()` and `fill_text()` all speak user space. Metrics read from a zoomed canvas came back inflated: at a 2.35x zoom the 0.1-quantized scale is 2.3, so Amiri's 23.8px subscript size (34px font) was reported as 54.7px, and sub/superscript runs sized from `subscript_size()` grew with the zoom instead of staying proportional to the run's font size.

Fix: measure at the paint's font size directly, matching the space every sibling API reports. Since font metrics scale linearly with size, this is exactly the old value divided back by the internal scale — but stable under any transform.

The regression test pins the metrics under a camera-style zoom transform (translate * scale * translate), under a device pixel ratio of 2, and against the `TextContext`-level API; it fails against the previous behavior (`ascender 99.28 should equal identity-transform value 43.16`).

Verified with an offscreen wgpu harness rendering RTL Arabic (Amiri and Courier New) with OS/2-metric sub/superscripts across an 8-step zoom ladder against Chromium 149 and Firefox 154 rendering SVG `<text direction="rtl">` built from the same OS/2 values: with this fix femtovg's max pixel diff vs either browser (6.6% / 4.7%) sits inside the browsers' own cross-engine disagreement (6.9%). Full test suites pass for textlayout, all-features, swash and no-default-features; `cargo fmt` and clippy are clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,15 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Fixed `Canvas::measure_font()` scaling its result by the canvas transform's
+  internal glyph-rasterization scale and the DPI factor. It now reports
+  user-space metrics that depend only on the paint's font size, matching
+  `measure_text()`, `TextContext::measure_font()` and the coordinate space
+  `fill_text()` consumes. Previously, metrics read from a zoomed canvas came
+  back inflated - for example sub/superscript runs sized via
+  `subscript_size()` grew with the zoom level instead of staying proportional
+  to the run's font size.
+
 - Fixed `Path::rounded_rect()` and `rounded_rect_varying()` flattening corners
   into ellipses when a radius did not fit. Radii that overlap along a side are
   now reduced by one common factor, as CSS Backgrounds and Borders Level 3 and
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1757,13 +1757,24 @@ where
             })
     }
 
-    /// Returns font metrics for a particular Paint.
+    /// Returns font metrics for a particular Paint, in user-space units.
+    ///
+    /// The values scale with the paint's font size only — the canvas
+    /// transform and DPI factor do not affect them, matching the space
+    /// [`measure_text`](Self::measure_text) reports and
+    /// [`fill_text`](Self::fill_text) consumes.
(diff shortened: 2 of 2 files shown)


matthargett:
Credit where due: @tronical's review comment on #291 (https://github.com/femtovg/femtovg/pull/291#discussion_r3783244858) asking for alignment/RTL coverage nudged us to test RTL more heavily — an 8-step zoom comparison of Arabic text with sub/superscripts against Chromium's and Firefox's SVG rendering. That's what surfaced this inconsistency (script runs sized from a zoomed canvas ballooned with the zoom), so we covered this additional case as a result in the same PR: fix plus a regression test that pins `measure_font` under zoom, hidpi and the `TextContext`-level API.

matthargett:
Visual demonstration — 8-step zoom ladder (0.6x–2.35x, camera zoom about the canvas center), RTL Arabic in Amiri (variable-width) and Courier New (fixed-width) with underline+strikethrough, Canvas-semantics drop shadows, and sub/superscripts sized from each font's OS/2 metrics read via `measure_font` at draw time, i.e. from the already-zoomed canvas:

RTL zoom comparison: femtovg master vs this fix vs Chromium vs Firefox

Top row: femtovg master (no fix), femtovg + this PR, Chromium 149, Firefox 154 — the browsers render SVG `<text direction="rtl">` built from the same OS/2 values. Bottom row: red diff overlays. Without the fix the subscript balloons as the zoom grows (the metrics pick up the internal quantized rasterization scale, e.g. 2.3 at 2.35x) and the superscript drifts out of place; with the fix femtovg's diff vs either browser (6.6% / 4.7% at max zoom) sits inside the 6.9% the two browsers disagree with each other.

tronical (approved):

tronical:
Unfortunately I can't push to resolve the conflict (maintainers aren't allowed to make edits on this PR).

tronical:
Rebased and pushed directly as commit c13efec93bcd7c116cada26f29409fc4a5d527a5

PR #318, opened 2026-08-16
Follow the first strong character for the bidi paragraph base direction

matthargett:
The bidi paragraph level was pinned left-to-right (`BidiInfo::new(text, Some(Level::ltr()))`), so a native RTL paragraph mis-ordered: an LTR word embedded between two Arabic words did not swap sides with its neighbors, and trailing neutral punctuation landed at the visual right end instead of the left. This resolves the base level per UAX #9 rules P2/P3 — first strong character wins, LTR when the text has none — which is what browsers do for `dir=auto` / `unicode-bidi: plaintext`, and what SVG renders for `direction="rtl"` text.

A second fix rides along because the first exposes it: the shaped-word cache keyed on `(size, word, fonts, variations)` but **not the run direction**. The same word shapes differently right-to-left (mirrored brackets per UAX #9 L4, reversed cluster order), and direction-neutral words — digits, `(yes)` — genuinely occur in both directions. Shaping `(yes)` in an LTR sentence and then in an Arabic sentence replayed the cached unmirrored LTR shaping inside the RTL run. The cache id now carries the direction bit.

### Stacking / merge order

This branch is **stacked on #316** (`measure_font` user-space metrics) — please merge #316 first; this PR then rebases to just the two bidi commits. The stacking is deliberate: the zoom-ladder cross-check below reads font metrics from a zoomed canvas and needs #316 to be meaningful.

### Tests (each verified fail-before / pass-after where applicable)

- `paragraph_base_direction_follows_first_strong_character` — RTL run ordering around an embedded LTR word, trailing `!` at the visual left, European digits as an LTR sequence inside an RTL paragraph, and an LTR-first control paragraph that keeps its old ordering. Fails against the pinned-LTR base.
- `brackets_mirror_in_rtl_runs_even_after_ltr_caching` — primes the cache with LTR shapings, then asserts `(` renders with the `)` glyph inside an Arabic sentence. Fails against the direction-less cache key.
- `bidi_browser_regression_cases` — cases distilled from shipped Firefox/WebKit bidi bugs: strong-less text stays LTR (Firefox 726420 class), an Arabic-Indic date joined by `/` stays one logical-order number run (UAX #9 W4, Firefox 459035), a bracket pair around an embedded LTR word resolves to the paragraph level together (UAX #9 N0, the Firefox 1177350 `(GMT+0800 (CST))` class), and Hebrew niqqud rides its base consonant through RTL reversal (Firefox 721821 class).
- `bidi_controls_are_invisible_and_zero_width` — FSI/PDI and LRM/RLM add no advance (the Firefox 1439018 isolates-as-tofu class).
- RLO directional override reverses Latin text visually with zero-advance marks — the WPT `2d.text.draw.fill.rtl` case from Chromium's 2025 CanvasTextNG bidi rework (their uniform-LTR fast glyph path shipped this broken; Google Docs wraps canvas strings in these overrides).

The last two lock in behavior the unicode-bidi + rustybuzz stack already gets right once the base direction and cache key are fixed — they pin it against regressions rather than fix new bugs.

### Cross-check vs Chromium and Firefox

Four mixed-direction paragraphs (embedded LTR word, trailing punctuation, digits, LTR-first control) across an 8-step zoom in/out ladder against both browsers rendering the same text as SVG (Amiri for both sides). The before panel misorders every RTL row; with the fix the run ordering matches the browsers exactly at every zoom step, and the remaining diff is glyph-rasterization noise, stable across the ladder:

bidi base direction before/after vs Chromium and Firefox across the zoom ladder

Full suites pass for textlayout, all-features, and swash; `cargo fmt` and clippy are clean. No allocation or per-frame cost change: the same bidi pass runs, only its paragraph-level input differs, and the cache key grows by one bool.

### DIFF
--- a/src/text/textlayout.rs
+++ b/src/text/textlayout.rs
@@ -42,6 +42,11 @@ pub(super) struct ShapingId {
     word_hash: u64,
     font_ids: [Option<FontId>; 8],
     variation_hash: u64,
+    // The same word shapes differently right-to-left than left-to-right
+    // (mirrored brackets, reversed cluster order), so the cached shaping is
+    // only valid for the direction it was produced under. Direction-neutral
+    // words (digits, punctuation) genuinely occur in both.
+    rtl: bool,
 }
 
 impl ShapingId {
@@ -51,6 +56,7 @@ impl ShapingId {
         word: &str,
         max_width: Option<f32>,
         variations: &FontVariations,
+        rtl: bool,
     ) -> Self {
         let mut hasher = FnvHasher::default();
         word.hash(&mut hasher);
@@ -63,6 +69,7 @@ impl ShapingId {
             word_hash: hasher.finish(),
             font_ids,
             variation_hash: variations.hash(),
+            rtl,
         }
     }
 }
@@ -239,6 +246,9 @@ pub fn shape(
         text,
         max_width,
         variations,
+        // The run cache keys the whole string; per-run direction is derived
+        // from the text itself below, so it needs no separate key bit.
+        false,
     );
 
     if !context.shaping_run_cache.contains(&id) {
@@ -282,7 +292,11 @@ fn shape_run(
         final_byte_index: 0,
     };
 
-    let bidi_info = BidiInfo::new(text, Some(unicode_bidi::Level::ltr()));
+    // The paragraph base direction follows the first strong character
+    // (UAX #9 rules P2/P3) rather than being pinned left-to-right: an Arabic
+    // or Hebrew sentence is an RTL paragraph, so its neutral punctuation
+    // sits at its visual left end and embedded LTR words order correctly.
+    let bidi_info = BidiInfo::new(text, None);
 
     // this controls whether we should break within words
     let mut first_word_in_paragraph = true;
@@ -313,7 +327,8 @@ fn shape_run(
         let mut byte_index = run.start;
 
         for mut word_txt in sub_text.split_word_bounds() {
-            let id = ShapingId::new(font_size, font_ids, word_txt, max_width, variations);
+            let is_rtl_run = hb_direction == rustybuzz::Direction::RightToLeft;
+            let id = ShapingId::new(font_size, font_ids, word_txt, max_width, variations, is_rtl_run);
 
             if !context.shaped_words_cache.contains(&id) {
                 let word = shape_word(
@@ -359,7 +374,14 @@ fn shape_run(
                             }
 
                             let subword_txt = &word_txt[..bytes_included];
-                            let id = ShapingId::new(font_size, font_ids, subword_txt, Some(max_width), variations);
+                            let id = ShapingId::new(
+                                font_size,
+                                font_ids,
+                                subword_txt,
+                                Some(max_width),
+                                variations,
+                                is_rtl_run,
+                            );
                             if !context.shaped_words_cache.contains(&id) {
                                 let subword = shape_word(
                                     subword_txt,
--- a/src/text/textlayout.rs
+++ b/src/text/textlayout.rs
@@ -45,8 +45,12 @@ pub(super) struct ShapingId {
     // The same word shapes differently right-to-left than left-to-right
     // (mirrored brackets, reversed cluster order), so the cached shaping is
     // only valid for the direction it was produced under. Direction-neutral
-    // words (digits, punctuation) genuinely occur in both.
-    rtl: bool,
+    // words (digits, punctuation) genuinely occur in both. `None` marks a
+    // whole-run key, where the direction is derived from the text itself.
+    rtl: Option<bool>,
+    // Letter spacing is baked into the cached advances by `shape_word`, so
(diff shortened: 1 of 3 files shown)


tronical:
I think the approach is correct. But this needs a rebase. I'm good with taking this in.

Overall, for the textlayout feature, I think it would be best to replace the implementation (that depends on unmaintained crates like ttf-parser and rusty buzz) with parley/fontique. That's maintained and well implemented.

src/text/textlayout.rs
```diff
@@ -313,7 +327,8 @@ fn shape_run(
         let mut byte_index = run.start;
 
         for mut word_txt in sub_text.split_word_bounds() {
-            let id = ShapingId::new(font_size, font_ids, word_txt, max_width, variations);
+            let is_rtl_run = hb_direction == rustybuzz::Direction::RightToLeft;
```
tronical: I suppose this could be moved outside of the loop?

src/text/textlayout.rs
```diff
@@ -313,7 +327,8 @@ fn shape_run(
...
+            let id = ShapingId::new(font_size, font_ids, word_txt, max_width, variations, is_rtl_run);
```
tronical: unrelated to this PR, but I realise that the shaping id is missing the letter spacing...

src/text/textlayout.rs
```diff
@@ -239,6 +246,9 @@ pub fn shape(
         text,
         max_width,
         variations,
+        // The run cache keys the whole string; per-run direction is derived
+        // from the text itself below, so it needs no separate key bit.
+        false,
```
tronical: Perhaps an Option<bool> would be sort of the more honest way of encoding this state :)

tronical:
Rebased and pushed directly as c13efec93bcd7c116cada26f29409fc4a5d527a5

tronical:
Oops, wrong PR closed. Sorry :)

src/text/textlayout.rs
matthargett: Done — hoisted out of the word loop (it's per run, not per word).

src/text/textlayout.rs
matthargett: Good catch, and it turned out to be a real collision, not just a hygiene point: `shape_word` bakes the spacing into the cached advances, so measuring the same word at spacing 0 and then at spacing 6 replayed the first shaping's advances. Since this PR was already reshaping `ShapingId`, the fix is included here: the key now carries a quantized letter-spacing field, pinned by `letter_spacing_does_not_collide_in_the_shaping_cache`, which fails against the previous key.

src/text/textlayout.rs
matthargett: Agreed — done. `rtl` is now `Option<bool>`: `Some(direction)` for word-level keys, `None` for the whole-run key where direction is derived from the text itself.

matthargett:
Rebased onto current master — the stacked `measure_font` commit dropped out now that it landed as c13efec, so this is down to the bidi commits plus the review fixes (run-level hoist, `Option<bool>` key encoding, and the letter-spacing cache key you spotted — that one was a live collision, details in the thread). Full suites pass on textlayout/all-features/swash, fmt and clippy clean.

On parley/fontique for the textlayout backend: agreed it's the right direction — parley gets bidi, script itemization, and font fallback from actively-maintained crates rather than our hand-rolled pipeline. Worth its own tracking issue given the API surface it touches; the browser-regression tests this PR adds are backend-agnostic on purpose (they assert glyph positions and glyph ids through the public API), so they'd carry over as the conformance harness for that migration.

matthargett:
@tronical do let me know if you want me to go ahead with the dependency change up in this PR. I did a dry run locally and it's doesn't seem too bad, but it will make this PR larger. Totally up to you as to how you want to stage which changes in which releases!

tronical (approved):

tronical:
> Rebased onto current master — the stacked `measure_font` commit dropped out now that it landed as [c13efec](https://github.com/femtovg/femtovg/commit/c13efec93bcd7c116cada26f29409fc4a5d527a5), so this is down to the bidi commits plus the review fixes (run-level hoist, `Option<bool>` key encoding, and the letter-spacing cache key you spotted — that one was a live collision, details in the thread). Full suites pass on textlayout/all-features/swash, fmt and clippy clean.

Thanks :)

> On parley/fontique for the textlayout backend: agreed it's the right direction — parley gets bidi, script itemization, and font fallback from actively-maintained crates rather than our hand-rolled pipeline. Worth its own tracking issue given the API surface it touches; 

Yeah, that wasn't a suggestion for this PR :)

tronical:
> @tronical do let me know if you want me to go ahead with the dependency change up in this PR. I did a dry run locally and it's doesn't seem too bad, but it will make this PR larger. Totally up to you as to how you want to stage which changes in which releases!

Dependency change towards parley? Nah, not in this PR :)

PR #319, opened 2026-08-17
Use RED for Gray texture sub-uploads

frantisekbohacek:
Hey, I've encountered an issue when using FemtoVG on a device with llvmpipe. That implementatoin seems to be too strict and not allow the internal R8 format for the pixel format passed to tex_sub_image_2d. From the docs it seems to me the correct format to actually pass is the RED format (despite the name, it just means it's one component per pixel, not that the pixel would have to be red)

Pass GL_RED, not the sized internal format GL_R8, as the external format argument of glTexSubImage2D. A sized format there is invalid on GLES3 and desktop GL and raises GL_INVALID_ENUM, so strict drivers silently drop every Gray texture update. Texture creation correctly uses R8 as the internal format and stays unchanged.

### DIFF
--- a/src/renderer/opengl/gl_texture.rs
+++ b/src/renderer/opengl/gl_texture.rs
@@ -203,7 +203,7 @@ impl GlTexture {
 
         match src {
             ImageSource::Gray(data) => unsafe {
-                let format = if opengles_2_0 { glow::LUMINANCE } else { glow::R8 };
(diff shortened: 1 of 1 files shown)


tronical (approved):
Good catch - interesting that this slipped for so long, but it seems that's indeed not valid (R8 isn't).

PR #320, opened 2026-08-17
Add two-point radial gradients and gradient transforms (createRadialGradient / gradientTransform)

matthargett:
Adds the two gradient capabilities Canvas 2D and SVG have that femtovg could not express, building on #314's elliptical radials:

- **Two-point (independently centered) radial gradients** — `Paint::two_point_radial_gradient[_stops]`, the general Canvas `createRadialGradient(x0, y0, r0, x1, y1, r1)` form and the geometry SVG `fx/fy/fr` reduces to. Concentric radials keep using the cheaper box-gradient path; the new shader variants (indices 11/12) engage only for genuinely offset centers.
- **Gradient transforms** — `Paint::set_gradient_transform` / `with_gradient_transform`, the role SVG's `gradientTransform` plays: a transform on the *paint* that the *path* does not receive. #314's elliptical squash is the axis-aligned special case of this; the composition order is defined and documented (caller transform wraps the whole gradient definition, elliptical squash included).

### Why it cannot be worked around

usvg — the crate femtovg's own SVG example uses — normalizes **every** radial gradient to a unit circle at the origin plus a matrix (the idiom design tools emit), and real-world radials carry non-uniform scales. A femtovg consumer today must collapse them to concentric circles. Linear gradients survive (an affine image of a linear gradient is another linear gradient, and usvg folds the transform into the endpoints) — radials cannot be folded.

### Solver correctness (guards the bugs other engines shipped)

The two-circle solver follows Skia's `SkConicalGradient` reference math: largest root with non-negative radius wins per Canvas paint order, `disc < 0` fragments are transparent (the anti-NaN guard, Firefox bug 1638051's class), the leading-coefficient-zero case falls back to the linear solution, and there is **no** focal clamp (SVG 1.1's clamp was removed in SVG 2 to match Canvas). Coincident circles follow SVG. The WPT `2d.gradient.radial.inside*/outside*` tests are concentric and pass on a concentric-only implementation — the discriminating tests here are non-concentric by design.

### Validation — synthetic scenes, both browsers, zoom ladder

Four cases (Babylon-style offset centers, a focal cone, a 35°-rotated ellipse no axis-aligned API can express, and a multi-stop focal radial through the LUT shader) against Chromium 149 and Firefox 154 rendering SVG `fx/fy/fr` + `gradientTransform`. Master's nearest expressible rendering is 5–22% of pixels wrong across an 8-step zoom in/out ladder; this PR renders **0.00% different from both browsers at every step**, matching the browser-vs-browser envelope exactly:

static: master best effort vs this PR vs Chromium vs Firefox

zoom ladder, synthetic scenes

### Validation — in-the-wild SVG logos

The Reddit mark (8 radial `gradientTransform`s) and the WebKit mark (linear gradients), rendered through a usvg-driven harness, against both browsers rendering the SVGs directly. Reddit's shading needs this PR: master's best effort is ~5% of pixels wrong at a 20/255 threshold, dropping to ~2.5–4.4% with the PR — the remainder is antialiasing plus Reddit's `clip-path` (arbitrary-path clipping is femtovg's largest remaining SVG gap and out of scope here). WebKit's linear-only gradients render identically before and after — the honest negative control, since usvg folds linear transforms into endpoints:

static: real-world logos

zoom ladder: real-world logos

### Budget and back-compat

- No uniform-array growth: the two-point parameters reuse the box-gradient `extent`/`radius`/`feather` slots under a dedicated shader variant; one `Transform2D` per gradient paint, one extra matrix multiply per draw call rather than per fragment.
- Serde: the new per-flavor `transform` field defaults to identity for older serialized paints, and radii serialized as scalars (the pre-#314 format) deserialize as circles through an either-shape helper — both pinned by tests. (The scalar-radii guard also covers #314's own format change.)
- 17 test suites pass across the feature matrix; `cargo fmt` and clippy clean (no new lib warnings vs master).


matthargett:
Pushed 324122e: the gradient transform now folds into a linear gradient's endpoints instead of composing onto the paint matrix.

Why: design-tool SVGs author linear gradients with endpoints a fraction of a unit apart under a `gradientTransform` whose translation is 1e5..1e6 — Mozilla's Firefox logo (`splash-logo.svg`, [file](https://raw.githubusercontent.com/rebeckerspecialties/femtovg/demo-assets/splash-logo.svg)) carries `matrix(480 0 0 -496 428684 1098420)` on a 0.43-unit span. Composed the old way, nanovg's `large = 1e5` offset rode through that transform and the paint-matrix translation reached ~1e7, where f32 resolves whole units: `t` quantized into stairs (left) along the gradient's iso-lines. Folding keeps the paint matrix at the global transform only (right):

stairs

An affine map keeps a linear gradient linear, so the fold is exact: the start maps directly and the end goes along the transformed iso-line normal (A^-T v) at the distance that keeps t = 1. For similarities that is just mapping both endpoints; for anisotropic or skewed transforms it is the only exact answer — pinned by a scale(2,1) test with hand-computed iso-lines. The huge-translation test fails without the fold (worst channel delta 74 vs ≤3 after).

(The *hole* in that flame is a separate, pre-existing LUT bug — ramps weren't padded past the last stop — fixed in #329 against master.)

src/renderer/params.rs
```diff
@@ -265,3 +320,30 @@ impl Params {
...
+
+/// Maps a linear gradient's endpoints through `transform` so the resulting
+/// user-space gradient is exactly the transformed one: the start point maps
+/// directly, and the end point is placed along the transformed gradient
+/// direction (A^-T v, the image of the iso-line normal) at the distance that
+/// keeps t = 1 there. For similarities this is just mapping both endpoints;
+/// for anisotropic or skewed transforms it is the only exact answer.
+/// Degenerate inputs (zero-length gradient, singular transform) fall back
+/// to mapping both endpoints.
+fn fold_linear_gradient_transform(sx: f32, sy: f32, ex: f32, ey: f32, transform: &Transform2D) -> (f32, f32, f32, f32) {
```
tronical, line 332: Could this be written in terms of Transform2D's `transform_point` perhaps, to simplify this code a little or make it more readable? :)

src/renderer/wgpu/shader.wgsl
```diff
@@ -223,6 +232,101 @@ fn conicAngleFraction(vertex: VertexOutput, params: Params) -> f32 {
...
+            let t_lo: f32 = min(root_a, root_b);
+            let t_hi: f32 = max(root_a, root_b);
+            // Canvas draws circles from the largest offset toward the smallest and
+            // does not overpaint, so the largest offset whose interpolated circle
+            // has a non-negative radius is the one that claims the fragment.
+            if (r0 + t_hi * dr >= 0.0) {
+                result.t = clamp(t_hi, 0.0, 1.0);
+                result.covered = true;
+            } else if (r0 + t_lo * dr >= 0.0) {
+                result.t = clamp(t_lo, 0.0, 1.0);
```
tronical: This is duplicate a few times. Perhaps it would be simpler to just set `result.t` unclamped and do the clamping once at the return site further down.

src/renderer/wgpu/shader.wgsl
```diff
@@ -223,6 +232,101 @@ fn conicAngleFraction(vertex: VertexOutput, params: Params) -> f32 {
...
+        if (a < 0.0) {
+            // One circle strictly contains the other, so every fragment is on
+            // some interpolated circle and the discriminant is never really
+            // negative. At the focal point it is mathematically zero and rounds
+            // just below, which would punch a hole exactly where the first stop
+            // belongs, so hold it at zero rather than reading it as a miss.
+            disc = max(disc, 0.0);
+        }
+        if (disc >= 0.0) {
+            let s: f32 = sqrt(disc);
```
tronical: Perhaps this could be written as `sqrt(max(disc, 0.0))` and then you can make `disc` immutable?

src/renderer/opengl/main-fs.glsl
```diff
@@ -44,6 +44,8 @@ varying vec2 fpos;
  #define SHADER_TYPE_FillGradientConic 8
  #define SHADER_TYPE_FillImageGradientConic 9
  #define SHADER_TYPE_FilterImageColorMatrix 10
+ #define SHADER_TYPE_FillGradientRadial 11
```
tronical: Suggest to call this FillGradientTwoPointRadial. Could also be called conical?

(same applies to wgsl)

tests/gradient_paint_conformance_wgpu.rs
```diff
@@ -0,0 +1,920 @@
...
+        required_limits: wgpu::Limits::downlevel_defaults().using_resolution(adapter.limits()),
+        experimental_features: wgpu::ExperimentalFeatures::disabled(),
+        memory_hints: wgpu::MemoryHints::MemoryUsage,
+        trace: wgpu::Trace::default(),
+    }))
+    .ok()?;
+    Some((device, queue))
+}
+
+fn render(device: &wgpu::Device, queue: &wgpu::Queue, draw: impl FnOnce(&mut Canvas<WGPURenderer>)) -> Vec<u8> {
```
tronical: These two functions are very much duplicated. I suggest to move them into a common module and share them among the tests.

tronical (approved):
Looks good overall, a few nitpicks inside

src/renderer/params.rs
matthargett: Done — the start and end points now go through `Transform2D::transform_point`; only the A^-T direction math still reads the matrix components.

src/renderer/wgpu/shader.wgsl
matthargett: Done — each branch sets `result.t` unclamped and it is clamped once before the return, which also brings the WGSL in line with the GLSL version.

src/renderer/wgpu/shader.wgsl
matthargett: Done — `disc` is immutable with `sqrt(max(disc, 0.0))`, and the containment case folds into the branch condition (`a < 0.0 || disc >= 0.0`) so a negative discriminant only counts as a miss for `a > 0`. Applied to the GLSL too.

src/renderer/opengl/main-fs.glsl
matthargett: Renamed to `FillGradientTwoPointRadial` / `FillImageGradientTwoPointRadial` (enum, GLSL, WGSL) and the shader functions to match. Kept "two-point radial" over "conical": femtovg already uses conic for the angular sweep (`conic_gradient`, Canvas `createConicGradient`), so "conical" next to it would read as that, and two-point radial is the term the Canvas `createRadialGradient` docs use for this form.

tests/gradient_paint_conformance_wgpu.rs
matthargett: Moved `headless_device()` and the render/readback into `tests/common/mod.rs` (the Cargo shared test module pattern) and switched this file to it. The other wgpu test files carry the same copies from before this PR; I left them alone here to keep the diff focused and will migrate them in a small follow-up once this lands.

tronical (approved):

PR #321, opened 2026-08-19
Execute image-filter lists as chains with CPU folding and two-scratch ping-pong

matthargett:
Executes a list of image filters as one chain — the model behind Canvas `ctx.filter` lists (`"blur(5px) brightness(1.2)"`) and SVG filter chains (the ubiquitous `feGaussianBlur → feOffset → feMerge` drop-shadow recipe) — as pure orchestration over the merged `filter_image` machinery (#317). No new shaders, no new uniform slots, no new dependencies.

### Real-world driver

This is the execution engine for a gap visible in stock assets today: the Google Workspace icon's arrow is a `mask` applied to a group carrying a `feFlood → feBlend → feGaussianBlur` filter chain. femtovg currently paints that group unmasked and unblurred — ~20–37% of pixels wrong depending on zoom:

Google Workspace icon: femtovg today vs Chromium

This PR alone does not close that gap — the group must first be captured into a layer (group-effects PR, next in the stack) — but the chain executor is the piece both the filter and the mask composite run through, and its contracts (borrowing, transient bounds) are shaped so that capture pass composes with no copies and no retained images.

### Why the folding is on the CPU

Folding two 4×5 matrices is an ~80-FLOP multiply that happens once per chain per frame; the alternative places are all worse: per-fragment in-shader composition would redo the fold millions of times per pass, and a wgpu compute prepass would cost a pipeline, a dispatch, and a buffer round-trip to save nanoseconds of CPU. Folding on the CPU is what keeps the GPU at **one pass** for any run of color operations. The GPU-side opportunity that *is* real — fusing a trailing color matrix into the blur's second internal pass, saving one full-screen pass in mixed chains — is a shader-level follow-up that this API doesn't constrain.

### API

`Canvas::filter_image_chain(target, &[ImageFilter], source)` plus `ImageFilter::fold_with()` (two color matrices compose into one by 4×5 multiplication, exact in unpremultiplied space — the space the shader applies matrices in, matching Skia's `asAColorMatrix`/`Compose` folding).

### Budget invariants (each held by a test, not a comment)

- **N adjacent color operations = one GPU pass, zero intermediates**: runs of color matrices fold on the CPU. Property tests pin fold == sequential application and constant-column ordering (`invert∘brighten ≠ brighten∘invert`).
- **Peak transient memory = 2× the source, regardless of chain length**: unfoldable passes ping-pong between at most two scratch images (a unit test asserts the transient count — the class behind WebKit bug 218422's ~600MB of per-primitive intermediates). The pass list is allocated once at exact capacity and `ImageFilter` is `Copy`, so building a chain never reallocates or deep-copies.
- **Scratches ride the flush lifecycle** through a new `acquire_transient_image` seam (the shadow-pass transient list, generalized). This is deliberately the interface a shared render-target pool can later drop into without touching any call site — shadows, chains, and future layer/mask captures would then share one bounded pool.

### Designed for what layers on top

The chain **borrows** `source_image`/`target_image` without taking ownership — only its internal scratches are flush-tied. That is the contract group effects need: a group-capture pass (SVG group opacity/filters/masks — see the Google Workspace icon case in our roadmap, whose arrow is a mask over a `feFlood→feBlend→feGaussianBlur` group) can render into its own transient, feed it in as `source_image`, and release on its own schedule — no copies, no extra retained images, and the same pool bounds total transient memory across features.

### Orientation contract

A color-matrix pass flips the stored image (the render-target convention documented on `filter_image`), the two-pass blur preserves it. The chain appends one identity pass when the flip count is even, so **every chain shape — including the empty list — ends in the same flipped-storage / sample-with-FLIP_Y convention** as a single color-matrix `filter_image` call. The cost (one extra full-screen pass, only for parity-even chains like a lone blur) is documented, with `filter_image` as the single-pass escape hatch; a test drives an asymmetric source through five chain shapes and asserts nothing mirrors.

### Fixed along the way: degenerate blur parameters

Writing the regression tests against shipped browser bugs caught a live bug here: `GaussianBlur { sigma: 0.0 }` divided the Gaussian coefficient by zero and **blanked the output** instead of passing the image through (Firefox bug 619968's class), and a huge sigma computed coefficients from the unclamped value while the shader loop used the clamped one (Firefox bug 441368's overflow class). Both backends now sanitize sigma at the one place coefficients are built; verified fail-before/pass-after.

### Regression classes covered (researched across Chromium/Skia, Firefox, WebKit, WPT)

| Class | Precedent | Test |
| --- | --- | --- |
| Fold == sequential, one pass per color run | Skia matrix folding | `folding_matches_sequential_application`, `chained_color_matrices_match_the_fold` |
| Peak memory bounded (2 scratches) | WebKit 218422 | `filter_chain_bounds_transient_images` |
| Pass order preserved | WebKit 220893's partial-chain class | `chain_order_is_preserved_across_pass_boundaries` |
| Alpha clamped between passes | Firefox 1577566 (alpha-amplify rendered white) | `alpha_amplifying_matrix_clamps_between_passes` |
| Orientation stable across shapes | the #285 FLIP_Y mirror class | `chain_orientation_is_stable_across_shapes` |
| sigma 0 / huge / NaN bounded | Firefox 619968, 441368; Skia's MAX_SIGMA precedent | `degenerate_blur_parameters_stay_bounded` |

Deliberately deferred and documented: sRGB-only operation (matches CSS filter functions; linearRGB `color-interpolation-filters` is a later slice), filter regions/subregions (femtovg filters are whole-image), and multi-input primitives (`feMerge` branches need more than two live buffers; the ping-pong invariant is stated for linear chains).

17 test suites pass across the feature matrix; fmt and clippy clean (no new lib warnings vs master).


matthargett:
Update: an effect-chain matrix cross-check against Chromium's `ctx.filter` (14 structured chains — order permutations, blur sandwiches, folding runs, degenerate parameters, semi-transparent content) caught a real bug in this PR and fixed it: the ping-pong scratches carried no `PREMULTIPLIED` flag, so consumers re-premultiplied their texels and **semi-transparent content darkened at every pass boundary** (a half-alpha green went gray through `opacity(0.5) blur(3px)` — 100% of pixels wrong). Single filters were exact, which is why the single-op validation missed it. Fixed by flagging the scratches (and documenting that chain targets need `PREMULTIPLIED | FLIP_Y`), pinned by a new test pushing half-alpha content through a two-pass chain.

After the fix the matrix reads: all pure color-matrix chains **0.00%** vs Chromium; every blur-bearing chain sits at 2–9%, fully accounted for by the pre-existing blur-kernel fidelity gap now filed separately with numbers. Also fixed: two stray harness examples had slipped into the tree and broke `cargo build --examples` on the CI matrix (the earlier red CI across the stack) — removed and propagated through #322/#323/#324.

matthargett:
heads-up on a ripple from #331 (zero-initialize GL textures at creation): `create_image_empty` on the GL backend now uploads a zeroed buffer, so texture creation is eager and O(width×height). `acquire_transient_image` here creates a fresh image per acquisition and frees at the next flush, so once #331 merges, filter/layer/mask-heavy frames pay that zero-init per scratch per frame on GL — a 1080p RGBA scratch is an ~8MB upload. wgpu is unaffected (it already guaranteed zero-init, and its tracking can elide the clear when the first use overwrites fully).

no change needed in this PR — this seam was shaped for a size-bucketed transient pool from the start, and pooling turns the zero-init back into a one-time cost per bucket. it does move the pool from "optimization" to "mitigation" on GL, so I'll bump it to the front of the queue once this stack lands.

tronical:
Looks generally sensible to have - good idea :). I'll let claude run a review over this.

src/image.rs
```diff
@@ -549,4 +549,129 @@ impl ImageFilter {
...
+    /// equivalent filter when both are color matrices.
+    ///
+    /// This is the load-bearing rule for filter chains: a run of N adjacent
+    /// color operations costs one GPU pass and zero intermediate textures,
+    /// because 4x5 matrices compose by multiplication on the CPU. The fold is
+    /// exact in unpremultiplied space - the same space the shader applies the
+    /// matrix in - matching how Skia folds via `asAColorMatrix`/`Compose`.
+    /// Returns `None` when either side is not a color matrix (a blur cannot
+    /// fold), leaving chain execution to run them as separate passes.
+    pub fn fold_with(&self, next: &ImageFilter) -> Option<ImageFilter> {
```
tronical: `ImageFilter` is `Copy` — which the PR description leans on — so this reads more naturally by value:

```rust
pub fn fold_with(self, next: Self) -> Option<Self>
```

Drops the `&` at both call sites and doesn't imply there's anything to borrow.

src/lib.rs
```diff
@@ -836,6 +836,115 @@ where
...
+    /// length; the scratches are freed at the next flush. All color work is
+    /// in unpremultiplied sRGB with output clamped to [0, 1] per pass, so an
+    /// alpha-amplifying matrix feeding a blur cannot blow out later passes.
+    ///
+    /// The target ends up in the same orientation convention as a single
+    /// color-matrix [`filter_image`](Self::filter_image) call: content stored
+    /// vertically flipped, sampled upright via [`ImageFlags::FLIP_Y`], and
+    /// carrying premultiplied alpha - create chain targets with
+    /// `ImageFlags::PREMULTIPLIED | ImageFlags::FLIP_Y` so semi-transparent
+    /// results composite once, not twice. An
```
tronical: Stray wrap mid-sentence: the line ends on `An` and continues `empty list degrades...` on the next. Reflowing the paragraph fixes it.

src/lib.rs
```diff
@@ -836,6 +836,115 @@ where
...
+            }
+            passes.push(*filter);
+        }
+
+        // Orientation parity: a color-matrix pass flips the image (the
+        // render-target convention), the two-pass Gaussian blur preserves it.
+        // Appending an identity matrix when the flip count is even pins the
+        // documented flipped-storage contract for every chain shape - and
+        // makes the empty list a copy.
+        let flips = passes
```
tronical: `ImageFilter` is `#[non_exhaustive]` (`src/image.rs:339`), so this `matches!` silently classifies any future variant as flip-preserving. Add a displacement map or an `feMorphology` later and the output mirrors, with no compile error and no test failure — which is out of step with how carefully the rest of this is pinned.

An exhaustive match forces the decision instead. `#[non_exhaustive]` doesn't block exhaustive matching inside the defining crate, so in `image.rs`:

```rust
pub(crate) fn flips_output(&self) -> bool {
    match self {
        Self::ColorMatrix { .. } => true,
        Self::GaussianBlur { .. } => false,   // two passes, two flips
    }
}
```

Then this becomes `.filter(|f| f.flips_output())` and adding a variant won't compile until someone says which it is.

src/lib.rs
```diff
@@ -836,6 +836,115 @@ where
...
+            .iter()
+            .filter(|f| matches!(f, ImageFilter::ColorMatrix { .. }))
+            .count();
+        if flips % 2 == 0 {
+            passes.push(ImageFilter::saturate(1.0));
```
tronical: `saturate(1.0)` is only the identity because `0.213f32 + 0.787f32`, `0.715 + 0.285` and `0.072 + 0.928` each happen to round to exactly `1.0`. True today, but that's an accident of the luminance constants rather than anything the function promises — and at the call site "saturate" reads as a real operation, not a no-op.

An explicit `ImageFilter::identity()` (or a `const IDENTITY_COLOR_MATRIX: [f32; 20]`) is exact by construction and says what the pass is for. `folding_respects_order_and_identity` in `image.rs` is relying on the same coincidence and would benefit too.

src/lib.rs
```diff
@@ -836,6 +836,115 @@ where
...
+        }
+
+        if passes.len() == 1 {
```
tronical: This early return is redundant — the loop below already handles it. With `passes.len() == 1` you get `last == 0`, so the first iteration takes the `i == last` branch, allocates no scratch, and makes the identical `filter_image(target_image, passes[0], source_image)` call.

Dropping it also lets `image_size` move into the allocation branch, so the single-pass path stops doing a lookup that `filter_image` immediately redoes, and the `.unwrap()` at line 940 goes away:

```rust
let mut scratch: [Option<ImageId>; 2] = [None, None];
let mut src = source_image;
let last = passes.len() - 1;
for (i, filter) in passes.iter().enumerate() {
    let dst = if i == last {
        target_image
    } else {
        match scratch[i % 2] {
            Some(id) => id,
            None => {
                let Ok((w, h)) = self.image_size(source_image) else { return };
                // Scratches hold premultiplied filter output; ...
                let Ok(id) = self.acquire_transient_image(w, h, ImageFlags::PREMULTIPLIED) else { return };
                scratch[i % 2] = Some(id);
                id
            }
        }
    };
    self.filter_image(dst, *filter, src);
    src = dst;
}
```

Same behaviour, about ten lines less. (A bad `source_image` still ends up a no-op, just via `filter_image`'s own early return rather than this one.)

src/lib.rs
```diff
@@ -4394,6 +4503,72 @@ fn layout_stores_the_run_baseline() {
     }
 }
 
+/// Chain execution stays within the femto budget: any run of color matrices
+/// folds to a single pass (zero transient images), and longer mixed chains
+/// ping-pong between exactly two transient scratches however long they get -
+/// the WebKit-600MB-intermediates class (bug 218422) made into an invariant.
+#[cfg(test)]
```
tronical: `#[cfg(test)]` is redundant here — `#[test]` items are already stripped from non-test builds, and none of the other ~30 tests in this file carry it.

src/renderer/opengl.rs
```diff
@@ -684,6 +684,12 @@ impl OpenGl {
         );
         blur_params.shader_type = ShaderType::FilterImage;
 
+        // Degenerate standard deviations must stay pass-through rather than
+        // poisoning the pass: sigma 0 (or negative / NaN) divides the Gaussian
+        // coefficient by zero and blanks the output, and a huge sigma must clamp
+        // to the same in-shader bound the loop uses so the coefficients and the
+        // iteration count agree. Near-zero renders as a visually exact copy.
+        let sigma = if sigma > 0.0 { sigma.min(8.0) } else { 1e-3 };
```
tronical: Two things around this clamp.

**The old clamp is now dead code.** `blur_params.image_blur_filter_sigma = sigma.min(8.)` at line 705 can never fire — `sigma` is already `<= 8.0` by the time it gets there. Worth deleting, along with the "GLES 2.0 does not allow non-constant loop indices" comment above it, which now describes a bound this line owns. Otherwise there are two places each claiming to enforce the 8.0 limit, which is the sort of drift that produced the bug you're fixing.

**The sanitization is duplicated, not centralized.** The PR description says both backends sanitize "at the one place coefficients are built", but it's two places, with this six-line comment copy-pasted into each (see `src/renderer/wgpu.rs:898`). The whole coefficient block is identical in both, so it extracts cleanly:

```rust
pub(crate) fn gaussian_blur_coefficients(sigma: f32) -> ([f32; 3], f32) {
    // ... the one copy of the comment ...
    let sigma = if sigma > 0.0 { sigma.min(8.0) } else { 1e-3 };
    let y = f32::exp(-0.5 / (sigma * sigma));
    ([1. / ((2. * PI).sqrt() * sigma), y, y * y], sigma)
}
```

Both backends then set the coeffs and `image_blur_filter_sigma` from one return value, and a third backend can't forget.

(Sanitizing further upstream is correctly ruled out, for what it's worth — `large_shadow_blur_records_unclamped_spec_sigma` at `src/lib.rs:3395` deliberately pins the unclamped sigma into the command. And `if sigma > 0.0` rather than `clamp` is the right idiom, since `f32::clamp` propagates NaN.)

src/renderer/wgpu.rs
```diff
@@ -890,6 +890,12 @@ fn gaussian_blur_filter(
     );
     blur_params.shader_type = ShaderType::FilterImage;
 
+    // Degenerate standard deviations must stay pass-through rather than
+    // poisoning the pass: sigma 0 (or negative / NaN) divides the Gaussian
+    // coefficient by zero and blanks the output, and a huge sigma must clamp
+    // to the same in-shader bound the loop uses so the coefficients and the
+    // iteration count agree. Near-zero renders as a visually exact copy.
+    let sigma = if sigma > 0.0 { sigma.min(8.0) } else { 1e-3 };
```
tronical: Same as `src/renderer/opengl.rs:692` — this clamp makes `blur_params.image_blur_filter_sigma = sigma.min(8.)` at line 910 dead, and the sanitization plus its six-line comment is now duplicated across both backends. A shared `gaussian_blur_coefficients(sigma)` helper collapses both copies.

tronical (approved):
Inline comments by claude :)

src/image.rs
matthargett: Done — `fold_with(self, next: Self) -> Option<Self>`; the `&` is gone at every call site (chain, tests).

src/lib.rs
matthargett: Done — reflowed the paragraph.

src/lib.rs
matthargett: Done — `ImageFilter::flips_output()` is an exhaustive match (`ColorMatrix` true, `GaussianBlur` false) and the parity count uses it, so a new variant will not compile until it declares its parity.

src/lib.rs
matthargett: Done — added `ImageFilter::identity()` (exact by construction) and used it for the parity pass and in `folding_respects_order_and_identity`; noted it in the changelog entry since it is new public API.

src/lib.rs
matthargett: Done — took your shape: no early return, and the size lookup plus the scratch allocation live in the `None` arm of the `match`, so the single-pass path allocates nothing and looks nothing up, and the `unwrap()` is gone.

src/lib.rs
matthargett: Done — removed the `#[cfg(test)]`.

src/renderer/opengl.rs
matthargett: Done — `renderer::gaussian_blur_coefficients(sigma) -> ([f32; 3], f32)` holds the one copy of the sanitization and its comment (including the GLES 2.0 constant-loop-bound reason); both backends copy the coefficients and set `image_blur_filter_sigma` from its result, and the dead `sigma.min(8.)` clamps are deleted. Kept `if sigma > 0.0` rather than `clamp` for the NaN reason you noted.

src/renderer/wgpu.rs
matthargett: Done — same helper as the GL side; the wgpu copy of the block and its dead clamp are gone.

tronical (commented):
Looks good to me. Let me toss Claude at it :)

src/image.rs
```diff
@@ -549,4 +562,140 @@ impl ImageFilter {
...
+    /// equivalent filter when both are color matrices.
+    ///
+    /// This is the load-bearing rule for filter chains: a run of N adjacent
+    /// color operations costs one GPU pass and zero intermediate textures,
+    /// because 4x5 matrices compose by multiplication on the CPU. The fold is
+    /// exact in unpremultiplied space - the same space the shader applies the
+    /// matrix in - matching how Skia folds via `asAColorMatrix`/`Compose`.
+    /// Returns `None` when either side is not a color matrix (a blur cannot
+    /// fold), leaving chain execution to run them as separate passes.
+    pub fn fold_with(self, next: Self) -> Option<Self> {
```
tronical, line 588: The fold composes the linear parts but drops the per-pass clamp, so it isn't equivalent to running the two filters when the first one can go out of range.

`ColorMatrix`'s own doc says the result "is clamped to `[0, 1]`", and both shaders do exactly that (`renderColorMatrix` → `clamp(vec4(r,g,b,a), 0, 1)` then re-premultiply). Folding removes that intermediate clamp, so whenever matrix A pushes a channel outside [0,1] and matrix B mixes or scales channels, the folded pass differs from two `filter_image` calls — and from a browser, where CSS/SVG clamp per filter function.

`brightness(2.0)` then `saturate(0.0)` on rgb `(0.8, 0.1, 0.1)`:

- sequential, clamped each pass: `1.0, 0.2, 0.2` → luma **0.370**
- folded: luma of `(1.6, 0.2, 0.2)` → **0.498**

~33 levels out of 255 — visible, and `filter: brightness(2) grayscale(1)` in Chromium gives the sequential value.

This isn't limited to hand-built matrices: `sepia`'s rows sum to ~1.35, so even the `sepia(0.8) → hue_rotate(1.1)` pair the tests use diverges on bright pixels. Skia isn't a precedent for unconditional folding either — `SkMatrixColorFilter` carries a `Clamp` flag and `appendStages` emits `clamp_01` per filter.

A cheap guard keeps the one-pass budget for nearly every real chain: fold only when the **first** matrix is range-safe, i.e. for each row `sum(positive coeffs) + constant <= 1` and `sum(negative coeffs) + constant >= 0` (worst case over inputs in [0,1]). `identity`, `opacity`, `grayscale`, `saturate(<=1)`, `hue_rotate` all pass; `brightness(>1)`, `contrast`, `sepia` fall back to a second pass. If you'd rather keep folding unconditional, the divergence should be stated in the `filter_image_chain` docs rather than described as exact.

Worth a line either way: alpha-zero pixels. Run sequentially, a matrix that zeroes alpha stores `(0,0,0,0)` and the next pass's `if (c.a > 0.0)` guard leaves it black; folded, the color survives into the second matrix.

---
_Generated by [Claude Code](https://claude.ai/code)_

tests/filter_chain_wgpu.rs
```diff
@@ -0,0 +1,290 @@
...
+        eprintln!("skipping: no wgpu adapter available");
+        return;
+    };
+    let src = solid(femtovg::rgb::RGBA8::new(180, 90, 40, 255));
+    let first = ImageFilter::sepia(0.8);
+    let second = ImageFilter::hue_rotate(1.1);
+    let folded = first.fold_with(second).expect("folds");
+
+    let chained = run_chain(&device, &queue, &src, &[first, second]);
+    let one_pass = run_chain(&device, &queue, &src, &[folded]);
```
tronical: Both sides of this comparison execute the same single GPU pass, so the assert can't fail except on GPU nondeterminism: `run_chain(&[first, second])` folds the pair inside `filter_image_chain` before any command is recorded, and `run_chain(&[folded])` is that same matrix.

To make it load-bearing, build the sequential side explicitly — `filter_image(scratch, first, src)` then `filter_image(target, second, scratch)` (two flips, so the orientation still lines up with the chain's) — and compare that against `run_chain(&[first, second])`. That version would also exercise the clamp question raised on `fold_with`.

Same gap on the CPU side: `folding_matches_sequential_application`'s `apply` helper doesn't clamp, which is precisely the step folding removes, so it can't catch it either.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/lib.rs
```diff
@@ -836,6 +836,109 @@ where
...
+    /// Applies a list of image filters as one chain, `filters[0]` first —
+    /// the execution model behind a Canvas `ctx.filter` list
+    /// (`"blur(5px) brightness(1.2)"`) and SVG filter chains.
+    ///
+    /// Runs of adjacent color-matrix filters fold into a single matrix on the
+    /// CPU (see [`ImageFilter::fold_with`]), so any number of consecutive
+    /// color operations costs one GPU pass. Passes that cannot fold ping-pong
+    /// between at most two transient scratch images sized like the source, so
+    /// peak transient memory is twice the source image regardless of chain
+    /// length; the scratches are freed at the next flush. All color work is
```
tronical: Peak transient memory is 3× the source, not 2×, for any chain containing a blur: `filter_image`'s blur path allocates its own full-size horizontal buffer for the duration of the pass (`images.alloc(self, source_image_info)` in `opengl.rs`, the `"blur horizontal"` texture in `wgpu.rs`) on top of the two scratches.

The invariant that matters — bounded, independent of chain length — still holds; it's just "two scratches plus one blur-internal buffer". Worth stating precisely here and in the PR description so the transient-pool sizing later targets the real peak.

---
_Generated by [Claude Code](https://claude.ai/code)_

tronical (commented):
Reviewed the chain executor, the fold, the parity rule, the sigma sanitization and the tests against both shader backends. The structure is good — `filter_image_chain` is pure orchestration over the existing `filter_image` machinery, the `shadow_images` → `transient_images` rename plus `acquire_transient_image` is a clean generalization of the lifecycle that already existed, and the ping-pong loop is correct (no self-aliasing; `passes` is never empty because the parity pass covers the even case, so the `passes.len() - 1` index is safe).

Things I checked and found correct:

- **Orientation parity.** Walked all five shapes: `[cm]` odd → target flipped; `[blur]` → identity appended → blur into scratch (unflipped), cm into target (flipped); `[cm, blur]` and `[blur, cm]` both land odd. Consistent with the single-`filter_image` convention. Scratches deliberately lacking `FLIP_Y` is right, since the filter shaders sample raw (`fpos.xy / extent`) and only the composite honours the flag.
- **`gaussian_blur_coefficients`.** The degenerate-sigma fix is real and the near-zero path is an exact passthrough, not an approximation: σ=1e-3 → `sampleCount = ceil(3e-3) = 1`, the loop breaks at `i >= 1`, so the output is `center * coeff.x / coeff.x`. NaN and negatives take the same branch. Clamping coefficients and the shader's σ to the same value is a genuine bug fix, and putting it in one place for both backends is the right shape.
- **Premultiplied scratches.** Matches what the shaders expect (`texType == 1` re-premultiplies on read); without the flag the per-pass darkening you described is exactly what would happen.

Three findings inline, one of them behavioural:

1. **`fold_with` drops the per-pass clamp** (`src/image.rs`) — the fold is exact for the linear part but not for the clamp that `ColorMatrix`'s own docs and both shaders specify. `brightness(2) → saturate(0)` comes out ~33/255 levels off from running the two filters separately, and off from Chromium. Suggested guard inline: fold only when the first matrix is range-safe.
2. **`chained_color_matrices_match_the_fold` is tautological** (`tests/filter_chain_wgpu.rs`) — the chain folds the pair before reaching the GPU, so both sides run the identical pass. The CPU-side property test can't catch the clamp divergence either, since its `apply` helper doesn't clamp.
3. **The 2× memory claim is 3× with a blur in the chain** (`src/lib.rs` docs) — the blur's internal horizontal buffer is live alongside the two scratches.

Minor, non-blocking:

- The two `let Ok(..) else { return }` bailouts inside the loop abandon the chain mid-way, leaving the target holding whatever it had before while earlier passes stay queued. Given `filter_image` is infallible-by-silence too it's consistent, but a caller can't distinguish "filtered" from "did nothing" — worth a doc sentence at least.
- `filter_image_chain(img, .., img)` (same source and target) reads and writes one texture in the folded single-pass case. Pre-existing for `filter_image`, but chains make it likelier to be tried; a doc note would help.
- The docs tell callers to create the target with `PREMULTIPLIED | FLIP_Y` but say nothing about the source: `renderColorMatrix` unpremultiplies unconditionally, ignoring `texType`, so a genuinely unpremultiplied source is divided by alpha it never had. Pre-existing from #317, not this PR's to fix, but the chain docs are the first place a user meets the requirement.
- Target sizing: chains inherit `filter_image`'s "target must match the source size" requirement (scratches are allocated at source size), which isn't stated on `filter_image_chain`.

CI is green across the matrix including `gpu-tests`. Nothing here blocks the design; #1 is the one I'd want resolved — either by guarding the fold or by documenting the divergence — before this becomes the execution path group effects build on.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/image.rs
matthargett: Fixed. `fold_with` now folds only when the first matrix is range-safe — every row's `const + Σ(positive coeffs) ≤ 1` and `const + Σ(negative coeffs) ≥ 0`, so its output stays in [0,1] and the clamp the fold drops was a no-op anyway. `brightness(>1)`, `contrast`, `sepia`, `saturate(>1)` therefore keep their own clamped pass, matching sequential `filter_image` and per-filter clamping in browsers/Skia; `next` may still overflow since it's clamped at the pass end. Added `color_matrix_range_safe`, a CPU unit test with the `brightness(2)`→`saturate(0)` witness, and a GPU test asserting the chain lands on the two-pass value, not the fold.

tests/filter_chain_wgpu.rs
matthargett: Fixed. The comparison now runs `run_chain` against a new `run_sequential` that executes each filter as its own `filter_image` pass (with the real per-pass clamp), so the two sides run different pass counts — renamed to `chained_color_matrices_match_sequential`. Added `range_unsafe_first_matrix_matches_sequential_not_fold` for the `brightness(2)`→`saturate(0)` case, which now folds to nothing and must match two passes. The CPU-side `folding_matches_sequential_application` clamps in its model and only asserts equality for range-safe first matrices.

src/lib.rs
matthargett: Fixed. The `filter_image_chain` docs and the CHANGELOG now state peak transient memory as twice the source image, or three times across a blur pass — the blur keeps its own full-size internal buffer alongside the two scratches. The bound is independent of chain length either way.

tronical (approved):

PR #322, opened 2026-08-20
Add layers: group capture with declared opacity and filter-chain effects

matthargett:
Adds layers: `Canvas::begin_layer(&LayerEffects)` / `end_layer()` capture a group of draws into a bounded offscreen image and composite it back with declared effects — **group opacity** as one fade over the finished layer (overlapping children no longer double-blend: the SVG group-opacity semantic and Canvas 2D's `beginLayer()`/`endLayer()` proposal) and/or an **image-filter chain** via `filter_image_chain` (#321). #323 (masks) and #324 (clip paths) stack on it.

### Evidence

**Google Workspace icon** (stock Google asset; a `feFlood→feBlend→feGaussianBlur` chain and a mask on one group). Without layers femtovg paints it unfiltered and unmasked: 19.7 % of pixels wrong at 1.6×. With this PR's capture + filter and #323's mask: **0.23 %** against Chromium and Firefox alike (their own disagreement: 0.02 %), inside that envelope at every zoom step (gif in #323).

Google Workspace icon: before, after, Chromium, Firefox, with diffs

**BuseyBench** (27 AI-generated SVG portraits; 169 opacity groups and 31 filtered ones in one file is typical). With this stack plus #336/#338/#339: **0.026 % mean structural difference** against both browsers at 460×260 (envelope 0.19 %). The group-shadow rule alone takes the mean 0.237 % → 0.205 %; the two gemini-3-1-pro-preview files, whose hair is an `feDropShadow` group, 0.35 % / 0.47 % → 0.03 % (third row).

group shadow and turbulence crops against both browsers

**The same corpus at 1080p** is what pulled the transient pool into this PR. gpt-6-astra, 0.000 % structural at 460×260, was **7.6 % / 3.1 %** of its box at 1920×1080: its 162 layers blew the 256 MiB transient budget and every layer past it silently lost its blur and opacity. With the pool: **0.18 % / 0.000 %**; corpus mean 9.9 % / 7.0 % → 1.5 % / 0.7 %; frames hold 20–57 MB (gpt-6-astra 29) where they requested 0.25–2.4 GB. Under a **48 MiB budget** (the Pi Zero guidance below) the corpus renders the same, 1 of its 2,373 layers passing through; 32 MiB costs 1.9 % / 0.9 %, 16 MiB 6.8 % / 4.5 %, every dropped effect reported. Under a 2× device scale the file still matches Chromium (0.20 % of the frame) holding 50 MB: the scaled scissor bounds the stores.

gpt-6-astra at 1080p: browsers, femtovg with and without the pool, diff maps

**WPT.** The web-platform-tests layer suite (`html/canvas/tools/yaml/layers.yaml`) is ported where the API can express it: `tests/wpt_layers_wgpu.rs`, 16 cases (global states, every Porter-Duff composite, nested and transformed layers, clip and blur across layer edges, css-filters, nested drop shadows, reset, non-invertible matrix, …), each drawn through layers and through the suite's own offscreen-canvas reference and compared per pixel. It settled the review's resize question and found two more semantics (below). SVGenius (100 icons, 12 with `opacity` groups): those groups through `begin_layer` are pixel-identical to folding the alpha.

### Design

- **Effects are declared at `begin_layer`** (Chromium's `beginLayer(filter)`), so the store is sized on creation: the scissor rect under any axis-aligned scale (a device-pixel-ratio transform still bounds it) plus the chain's blur reach (Gaussians compound in quadrature, 3·√Σσ² + 2), rounded up to 64 px, spanning the current render target when that is an image larger than the canvas. A rotated/rounded scissor falls back to the target size and keeps clipping per fragment.
- **A layer says whether it captured.** `begin_layer` returns `bool` under `#[must_use]`; `false` is a pass-through layer whose effects are dropped — over the transient budget, past the backend's texture limit (`Renderer::max_texture_size()`, queried from GL/wgpu; a VideoCore IV reports 2048), or degenerate bounds — with begin/end balanced either way. `filter_image_chain` returns `Result` (`ErrorKind::TransientImageBudgetExceeded`); a layer whose chain cannot get a scratch composites its unfiltered capture rather than a blank.
- **Orientation rides #321's contract** (flipped storage sampled via FLIP_Y, one parity flip per chain); a test pushes asymmetric content through a filtered layer and asserts nothing mirrors.
- **Rendering attributes reset inside a layer**: alpha (the outer alpha folds into the composite), the composite operation, and the shadow — the shadow state at `begin_layer` is cast once by the layer's *result*, the Canvas 2D layer rule and what `feDropShadow` on an SVG `<g>` means; set inside, it applies per draw. Frame boundaries follow WPT: an open layer keeps capturing across a same-size `set_size`, a resize or `reset()` discards open layers as a Canvas 2D reset does, and a layer under a non-invertible transform draws nothing.
- **Transient images are pooled within the frame** (`src/transient.rs`). A layer's capture and result return once its composite is recorded, a chain's scratches after its last pass, shadow and mask coverage after their draws; the next acquire of that size takes them (commands run in order, so no synchronization; the store is cleared on begin). Layer stores round to 64 px so siblings with different blur reaches share one; shadow coverage rounds to 8 px because shadows are many and small (600 glyph shadows: 650 images exact, 60 at 8 px for 18 % more fill, 8 at 64 px for 235 % more). Peak memory is nesting depth × layer size, not layer count. This belongs in the landing because the numbers are small: at 1080p a viewport-sized layer is 4.7 MB and a blurred one four images, so **13 blurred (57 plain) layers fill the 256 MiB budget and a Raspberry Pi Zero's 64–128 MB of GPU memory holds 3–6**, while the portraits above open 31–200 a frame.
- **`set_transient_image_budget()`** caps what is held at once; its docs size it for the platform (32–48 MiB on a Pi Zero, whose whole GPU share is 64–128 MB with 8 MB for the 1080p framebuffer). `transient_image_bytes()` reports what a frame held. Shadows draw through the same pool and budget.

### Tests

wgpu: no double-blend on overlap, filtered layer not mirrored, blur reaches past the edge and survives scissor edges, nested opacities multiply, composite honors the outer scissor, an open layer survives a flush and the frame's `set_size`, layers degrade past the budget and report it, the group shadow is cast once (overlapping discs under a 50 % shadow read 50 %, not 75 %), forty blurred siblings under a one-layer budget render every effect from cleared stores, the 16 WPT ports. Unit: store size under a scaled scissor and quadrature padding, tracked target across `set_size`, resize discards open layers, texture-limit pass-through, 8 px shadow stores, siblings reuse one store and nesting one per level, a one-layer budget fits two hundred layers, shadows and masks reuse their pairs, the chain-degrade path, unbalanced `end_layer` ignored. Full suite green across the feature matrix; fmt/clippy clean.


matthargett:

|---|---|---|
| `mask` | 15/27 | #323 |

| measure | value |
|---|---|

corpus at parity

filter integration

matthargett:
### 1080p: the transient-image budget is the first thing a real file hits (2026-09-07)

Measured on BuseyBench's `gpt-6-astra.svg` with the whole open stack (this PR's three, #334, #336, #338, #339) against Chromium 131 and Firefox DE, framed identically in a 1920x1080 frame with the 1024-unit SVG at 1080 px (harness and `make_ref.py` now take `FRAME_W/FRAME_H/BOX/BOX_X/BOX_Y`):

| framing | >20/255, % of SVG box | structural (2-px erosion) | mean delta | browser envelope |
|---|---|---|---|---|
| 460x260 (200 px box) | 0.45 % | 0.000 % | 1.8/255 | 0.03 % |
| **1920x1080 (1080 px box)** | **7.6 %** | **3.1 %** | 5.4/255 | 0.02 % |
| 1920x1080, `set_transient_image_budget(4 GiB)` | 0.18 % | 0.000 % | 1.4/255 | 0.02 % |

The file opens **162 layers in one frame** (opacity groups and `feGaussianBlur` shading), each sized to the 1080 px viewport scissor plus blur padding, about 5 MB apiece - some 800 MB requested against the 256 MiB default. Past the budget a layer degrades to pass-through, which drops its blur *and* its group opacity, so the shading blobs paint sharp and at full strength (right column of the sheet: red is what survives a 2-px erosion). Ablations on the reference side rule the other suspects out: capping the browsers' blur sigma at femtovg's 8 px moves the diff by 0.01 points, removing every stroke from both sides removes 2.3 points of ribbons and no structure.

gpt-6-astra at 1080p: browsers, femtovg with and without the budget, diff maps

3x cheek crops

Two conclusions for this PR's design:

- Sizing layers to their own bounds is not enough. With the harness scissoring each group to usvg's layer bounding box (filter region included, as a browser would), the sum for this file still exceeds the budget and the later, more visible layers are the ones that degrade - 10.5 % / 5.5 % structural, worse than viewport-sized. The lever is **reuse**: a layer's backing image is free for the next layer the moment its `end_layer` composite is recorded, because commands execute in order; nothing needs to wait for the flush or the GPU. Peak memory then is nesting depth x layer size (a handful of images here), not layer count x size. That is the roadmap's transient pool, and this measurement moves it from "optimization-only" to required for correctness at 1080p on a file of ordinary complexity.
- The degradation is silent. A frame that quietly loses its blurs and opacities should at least be observable - a counter or a `debug_assert`-free warning hook - so an integration can tell "the budget bit" from "the effect is missing".

Numbers and the ablation switches are in `harness/README.md` (section "Other framings: 1080p") on demo-assets.

matthargett:
Corpus-wide follow-up to the comment above, same framing (1080 px box in a 1920x1080 frame), every BuseyBench file, both browsers:

- **26 of 27 files request more than the 256 MiB budget per frame** with viewport-sized layers (250 MB–2.4 GB; 31–200 layers each) and **24 of 27 render differently because of it**. Corpus mean, % of the SVG box: **9.9 % / 7.0 % structural with the default budget, 1.5 % / 0.7 % with it lifted** (Firefox the same; envelope 0.15 %). The worst: qwen3-8-flash 69 % / 63 % → 1.1 % / 0.4 %, claude-opus-5 27 % / 25 % → 0.15 % / 0.00 %.
- **No file nests layers** (max depth 1), so a pool that frees a layer's images at `end_layer` peaks at **20–39 MB** across the corpus with today's viewport-sized layers (the largest single blurred layer: four images of (1080 + 2·pad)² × 4 B), 20–22 MB with bounding-box-sized ones — against the gigabytes requested per frame now. The model is `harness/poolsim.py` on a `LAYER_LOG=1` run; per-file numbers in `harness/busey-1080p-metrics.json` and `-summary.txt` on demo-assets.
- Ten files stay above 0.05 % structural with the budget lifted (gpt-5-2-pro 6.8 %, gpt-5-6-terra-pro 3.2 %, gemini-3-1-pro-preview-custom-tools 2.7 %, gpt-5-6-sol-pro 2.6 %, …) — separate 1080p-scale items, not attributed yet and not this PR's.

src/lib.rs
```diff
@@ -1724,19 +2118,42 @@ where
...
+            }
+        }
+        self.transient_images = kept;
+    }
+
+    /// After a flush the renderer starts the next command stream on the
+    /// screen target; if drawing is currently redirected (an open layer, or a
+    /// caller-selected image target), the redirect must be re-issued so the
+    /// next frame's commands keep landing where the canvas state says.
+    fn reissue_render_target_after_flush(&mut self) {
```
tronical, line 2184: This makes `current_render_target` load-bearing across frames, and that collides with `set_size()`, which pushes a raw `SetRenderTarget(Screen)` command **without** updating `current_render_target`:

```rust
// src/lib.rs:522
pub fn set_size(&mut self, width: u32, height: u32, dpi: f32) {
    ...
    self.append_cmd(Command::new(CommandType::SetRenderTarget(RenderTarget::Screen)));
}
```

Combined with "layers stay open across a flush", a layer open at a frame boundary silently loses the redirect. Every femtovg example calls `set_size()` at the top of each frame (`examples/demo.rs:169`), so this is the normal loop, not a corner case.

I reproduced it against this branch:

```rust
#[test]
fn probe_set_size_desyncs_open_layer_target() {
    let mut canvas = Canvas::new(RecordingRenderer::default()).unwrap();
    canvas.set_size(200, 200, 1.0);
    canvas.begin_layer(&LayerEffects::new().with_opacity(0.5));
    canvas.flush_to_output(());
    // Frame 2, exactly as every example starts a frame:
    canvas.set_size(200, 200, 1.0);
    let mut p = Path::new();
    p.rect(10.0, 10.0, 50.0, 50.0);
    canvas.fill_path(&p, &Paint::color(Color::rgb(255, 0, 0)));
    // ... dump SetRenderTarget commands and current_render_target
}
```

Output:

```
frame-2 target commands = [Image(ImageId(1v1)), Screen]
canvas thinks target     = Image(ImageId(1v1))
```

The reissue lands first, then `set_size` overrides it back to `Screen` while the tracked value still says `Image(L)`. So frame 2's red rect is painted straight onto the screen at full opacity, and `end_layer()` then composites a layer image that never received it — at 50 % over content already drawn opaque. No error, no assert; just wrong pixels, and the more layers a frame has the harder it is to trace.

`set_render_target` early-outs on equality (line 739), so any future `set_render_target(Image(L))` is also a no-op while the streams are out of sync — the desync is sticky, not self-healing.

Two things worth doing:

1. Make `set_size` go through `set_render_target(RenderTarget::Screen)` (or otherwise assign `current_render_target`) so the tracked target can never diverge from the stream. This is a latent bug for caller-selected image targets too, independent of layers — it just becomes reachable in the normal loop now.
2. `debug_assert!(self.layers.is_empty())` in `set_size`, or document that resizing with an open layer is undefined — a resize invalidates the store's device-space bounds anyway.

`tests/layer_effects_wgpu.rs:308` (`open_layer_survives_a_flush`) covers the flush but not a `set_size()` after it; adding that call to the test would pin this.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/lib.rs
```diff
@@ -916,32 +1101,239 @@ where
...
+                _ => 0.0,
+            })
+            .fold(0.0f32, f32::max);
+        let pad = if max_sigma > 0.0 {
+            (max_sigma * 3.0).ceil() + 2.0
+        } else {
+            0.0
+        };
+
+        let (keep_scissor, rect) = match state.scissor.as_rect(canvas_w, canvas_h) {
```
tronical: `Scissor::as_rect` bails out on *any* scale in the CTM captured at `scissor()` time:

```rust
// src/lib.rs:223
// Abort if we're scaling
if a != 1.0 || d != 1.0 {
    return None;
}
```

That's correct for its original callers (fast-path guards), but here `None` means "span the whole canvas". So on any canvas with a scale in the CTM — device-pixel-ratio, zoom, `scale()` for units — the scissor-derived bound quietly doesn't apply and every layer allocates a full-canvas store.

Measured on this branch, an 800×600 canvas with a 60×40 scissor:

```
canvas.scale(2.0, 2.0); canvas.scissor(50, 25, 60, 40)  ->  store 832x640
                        canvas.scissor(50, 25, 60, 40)  ->  store  64x64
```

170× the pixels. This is the headline memory rule in the PR description and the CHANGELOG ("sized to the current scissor rect plus the blur reach, not the whole canvas"), and it silently doesn't hold in what I'd expect to be the majority of real setups. It also feeds the budget-degradation path: at 1080p a full-canvas store is ~8.4 MB, so `set_transient_image_budget`'s default is reached at a nesting depth the pool was supposed to make irrelevant, and layers start passing through with no diagnostic.

The 1080p BuseyBench numbers in the description are the case I'd most want re-checked — if that harness draws at scale 1 with an explicit viewport, the reported 20–94 MB frames may not be what an app with `scale(dpr, dpr)` gets.

The fix looks small: the bail-out at `a != 1.0 || d != 1.0` is only needed by the fragment-mask fast paths, not by a caller that just wants a device-space bounding rect. With `b == 0 && c == 0` already established, an axis-aligned scale maps cleanly:

```rust
let hw = extent[0] * a.abs();
let hh = extent[1] * d.abs();
Some(Rect::new(x - hw, y - hh, hw * 2.0, hh * 2.0))
```

Either extend `as_rect` (and confirm the other callers tolerate a scaled rect), or give `begin_layer` its own scissor-bounds helper that handles scale. Worth a unit test at scale ≠ 1 either way — `layer_bounds_follow_the_scissor` and every wgpu test run at dpi 1.0, which is why this doesn't show up.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/lib.rs
```diff
@@ -916,32 +1101,239 @@ where
...
+            .map(|f| match f {
+                ImageFilter::GaussianBlur { sigma } if sigma.is_finite() && *sigma > 0.0 => sigma.min(8.0),
+                _ => 0.0,
+            })
+            .fold(0.0f32, f32::max);
+        let pad = if max_sigma > 0.0 {
+            (max_sigma * 3.0).ceil() + 2.0
+        } else {
+            0.0
+        };
```
tronical: The pad takes the **max** sigma over the chain, but successive Gaussians compound: two blurs of σ each have the reach of a single σ√2, and n of them σ√n. Reach combines in quadrature, so this under-pads any chain with more than one blur, and the layer's own content gets its blur tail clipped at the store edge (the thing the padding exists to prevent).

For a chain of three σ=8 blurs on a 200×200 scissor: `max` gives pad 26 → 252 → 256 px store, quadrature gives pad 3·13.86+2 ≈ 43.6 → 288 → 320 px. The 64-px rounding hides the two-blur case but not this one.

```rust
let sum_sq = effects
    .filters
    .iter()
    .filter_map(|f| match f {
        ImageFilter::GaussianBlur { sigma } if sigma.is_finite() && *sigma > 0.0 => Some(sigma.min(8.0)),
        _ => None,
    })
    .map(|s| s * s)
    .sum::<f32>();
let pad = if sum_sq > 0.0 { (sum_sq.sqrt() * 3.0).ceil() + 2.0 } else { 0.0 };
```

Reduces to the same value for the single-blur case, which is what the tests cover today. Nice touch that `sigma.min(8.0)` matches `gaussian_blur_coefficients`' own clamp (`src/renderer.rs:296`) — the shadow path at line 1983 doesn't, and over-pads for large sigma as a result. Not this PR's bug, but the layer path having the better version is worth noting.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/lib.rs
```diff
@@ -916,32 +1101,239 @@ where
...
+    /// Canvas 2D `beginLayer()` rule for its layer rendering attributes, and
+    /// what SVG's `feDropShadow` on a `<g>` means. Set the shadow state
+    /// again inside the layer to shadow individual draws as well. Layers nest;
+    /// each level costs one transient image (plus one more if filtered),
+    /// released at the next flush through the same pool seam the filter
+    /// chain and shadow passes use. When the backing store cannot be
+    /// allocated (degenerate or absurd bounds), the layer degrades to a
+    /// pass-through: drawing continues on the current target and `end_layer`
+    /// only rebalances state.
+    pub fn begin_layer(&mut self, effects: &LayerEffects) {
```
tronical: Graceful degradation is the right default, but as designed there is no way for a caller to learn it happened. `begin_layer` returns `()`, and a pass-through layer means the group opacity and the whole filter chain are silently dropped — the draws just appear unfaded and unfiltered. `ErrorKind::TransientImageBudgetExceeded` is added to the public enum but is unreachable from any public entry point: `begin_layer` swallows it via `.ok()` at line 1211 and `filter_image_chain` discards it with `let _ =` at line 1060.

`transient_image_bytes()` tells you the frame was near the cap but not *which* layer lost its effects, and a nested set can degrade partway through. For an app shipping this, "the blur disappeared on some machines at some window sizes" is a hard bug to chase from that signal alone.

Since this is the landing commit, it's cheap to make it observable now and expensive later:

```rust
#[must_use = "a false return means the layer's effects were dropped"]
pub fn begin_layer(&mut self, effects: &LayerEffects) -> bool
```

`bool` (or `Result<(), ErrorKind>`) still lets callers ignore it, keeps begin/end balanced either way, and gives the new `ErrorKind` variant a reachable path. Same for exposing a checked `filter_image_chain`. Either way the doc comment should say outright that effects are silently dropped past the budget — right now it reads as if only "degenerate or absurd bounds" trigger pass-through, and the budget case is documented over on `set_transient_image_budget` instead.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/lib.rs
```diff
@@ -836,20 +933,90 @@ where
...
+
+    /// Caps the memory held by transient images - layer backings, filter-chain
+    /// scratches, shadow coverage and mask coverage - at `bytes` (default
+    /// 256 MiB). Within a frame a layer's images are reused by the next layer
+    /// of the same size once its composite is recorded, so what counts against
+    /// the cap is the deepest nesting, not the number of layers: at 1080p a
+    /// viewport-sized layer is 4.7 MB and a blurred one four times that, so a
+    /// frame of hundreds of sibling layers stays at a few tens of MB. Once the
+    /// cap is reached, further layers degrade to pass-through, chains run
+    /// unfiltered, shadows and masks are skipped, rather than allocating
```
tronical: These docs describe masks ("mask coverage", "shadows and masks are skipped"), but masks are #323 — `grep -c 'begin_mask\|LayerMask' src/lib.rs` is 0 on this branch. Same in the `ErrorKind::TransientImageBudgetExceeded` doc in `src/error.rs:44` ("a layer backing, filter scratch or mask coverage").

Public API docs on a released crate shouldn't reference a feature that isn't there yet. Drop the mask mentions here and add them back in #323, so this PR's docs describe what this PR ships.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/lib.rs
```diff
@@ -1597,8 +1989,8 @@ where
         let maxx = (shape_bounds.maxx + pad).ceil();
         let maxy = (shape_bounds.maxy + pad).ceil();
 
-        let width = (maxx - minx) as usize;
-        let height = (maxy - miny) as usize;
+        let width = round_up_transient((maxx - minx) as usize);
+        let height = round_up_transient((maxy - miny) as usize);
```
tronical: Rounding the *shadow* offscreens to 64 px is a behavior change for existing shadow users, and the trade-off is less clearly a win here than it is for layers.

Pooling pays off when consecutive shadows share a size, which needs the rounding — agreed. But the cost isn't only the extra bytes: both backends allocate the horizontal-blur intermediate fresh from the source's dimensions on every blur pass and destroy it immediately (`opengl.rs:703` `images.alloc(...)` / `images.remove(...)`, `wgpu.rs:895` `create_texture`), and the two blur passes plus two `clear_rect`s all cover the rounded-up area. A 20×20 shadowed shape goes from a 24×24-ish offscreen to 64×64 — roughly 7× the fill and clear work per shadow, and one intermediate texture create/destroy either way.

Text is where I'd expect this to bite: many small, differently-sized shadowed glyph or icon draws per frame, each now paying the 64-px floor, and sizes varying enough that the pool hit rate is what decides whether it nets out.

Was this measured on a shadow-heavy scene, or only on the layer corpus? If it's a wash, the comment here should say the rounding is for pool sharing across shadows and that the extra fill was measured as acceptable. If it isn't, a smaller granularity for the shadow path (or `next_power_of_two` capped at 64) would keep the pooling win without the floor on small shapes.

---
_Generated by [Claude Code](https://claude.ai/code)_

tronical (commented):
This is a pretty cool API :). Some comments from claude within.

tronical (approved):
Maybe the transient image handling could go into a separate module and be encapsulated (`TransientImagePool` or so). It's starting to get a bit of code :)

I'll approve and trust you that you'll address the comments as you see fit (or ignore them ;-).

src/lib.rs
matthargett: Fixed in fe461f4. `set_size` now assigns the tracked target alongside the stream, and defines the open-layer cases from WPT: a same-size `set_size` at a frame boundary re-issues the layer's target (2d.layer.flush-on-frame-presentation), a resize discards open layers as a Canvas 2D reset does (2d.layer.reset), and `reset()` does the same. `open_layer_survives_a_flush` now calls `set_size` after the flush; two unit tests pin the tracked target and the discard.

src/lib.rs
matthargett: Fixed in fe461f4. `begin_layer` sizes its store from a `Scissor::device_bounds` helper that accepts an axis-aligned scale (`as_rect` keeps its bail-out for the blit fast path). `layer_bounds_follow_a_scaled_scissor` pins `scale(2)` + a 60×40 scissor to a 128×128 store; the 1080p corpus numbers in the description are re-measured at a 2× scale below.

src/lib.rs
matthargett: Fixed in fe461f4: the pad is `3·sqrt(Σσ²) + 2`, and the shadow path clamps its sigma at the shader's 8 like the layer path. `chained_blurs_pad_in_quadrature` pins three σ=8 blurs on a 200 px scissor to a 320 px store (max would have given 256).

src/lib.rs
matthargett: Done in fe461f4: `begin_layer` returns `bool` under `#[must_use]`, `filter_image_chain` returns `Result` (so `TransientImageBudgetExceeded` is reachable), and the `begin_layer` doc says outright when a layer passes through — over the budget, past the backend's texture limit, or degenerate bounds. The limit is new too: `Renderer::max_texture_size()`, queried from GL / wgpu, replaces the fixed 8192 guard (a VideoCore IV reports 2048), with a test that a store past it degrades rather than fails.


Done: `src/transient.rs` holds the pool (`TransientPool`: acquire with reuse, release, release-all at flush, the budget and the two granularities); `Canvas` keeps one field. Beyond the review: the WPT layer suite is now ported where the API can express it (`tests/wpt_layers_wgpu.rs`, 16 cases drawn through layers and through the suite's own offscreen-canvas references), which is what caught the reset/resize semantics and two more — a layer opened while rendering to an image larger than the canvas now spans that image, and one opened under a non-invertible transform draws nothing, as Canvas 2D specifies.

src/lib.rs
matthargett: Removed here in fe461f4; #323 adds the mask wording back where the feature lands.

src/lib.rs
matthargett: Not measured on a shadow-heavy scene before, so I modelled one: a frame of 600 glyph-sized shadowed draws (heights 9–48 px, σ 1–4). Exact sizes allocate 650 images per frame; 8 px rounding 60 for 18 % more fill; 16 px 26 for 46 %; 64 px 8 for 235 %. Shadows now round to 8 px and layers keep 64 (fe461f4, numbers in `SHADOW_GRANULARITY`'s comment); a unit test pins a 20 px shape at a 40×40 store. The per-pass blur intermediate you point at is pre-existing and worth pooling separately.

PR #323, opened 2026-08-20
Add layer masks (luminance / alpha) completing SVG group-effect rendering

matthargett:
Adds layer masks: `LayerEffects::with_mask(image, kind, x, y, w, h)` multiplies a captured layer's alpha by a mask image's coverage — `MaskKind::Luminance` (SVG `mask`'s default `mask-type`: the mask's luminance times its alpha, via the new `ImageFilter::luminance_to_alpha()`) or `MaskKind::Alpha`. The mask applies **after** the layer's filter chain, SVG's order for a group carrying both `filter` and `mask`. On master with #336 and #339 merged: a behavior-neutral helper refactor, a blit fix the mask work uncovered, the feature, and the fixes from a review round, each pinned by a test that failed before it.

### Evidence

**Google Workspace icon** (stock Google asset: the arrow is a `mask-type:alpha` mask whose content is a `gradientTransform` radial with fully opaque stops — so it exercises alpha coverage and the mask's placement, not luminance; the stop colours and the transform cannot change its coverage — on a group carrying a `feFlood→feBlend→feGaussianBlur` chain). Without layers and masks femtovg paints it unmasked and unblurred, 19.7 % of pixels wrong at 1.6×. With the layer stack and this PR, against Chromium 131 rendering the SVG directly, 460×260 pivot-zoom ladder:

| zoom | 0.6× | 0.9× | 1.0× | 1.3× | 1.6× | 1.9× | 2.35× |
| --- | --- | --- | --- | --- | --- | --- | --- |
| px > 20/255 | 0.05 % | 0.03 % | 0.06 % | 0.08 % | 0.06 % | 0.60 % | 4.4 % |
| structural (2 px erosion) | 0.000 % | 0.000 % | 0.000 % | 0.000 % | 0.000 % | 0.000 % | 3.0 % |

The two top steps sit past the blur shader's 8 px sigma cap, a GLES 2.0 bound on the 24-tap loop that predates this stack; the residual there is a narrower blur band, the angled ribbons in the animation's diff panels. #340, stacked on this PR, renders such blurs as quadrature passes and takes 2.35× to 0.06 %.

At 1.6× the stack differs from Chromium by 0.06 % of pixels and from Firefox by 0.04 %, against a 0.05 % browser-vs-browser envelope.

Google Workspace icon: master, current stack, Chromium 131, Firefox, with diffs

zoom ladder

**Luminance times alpha.** The first version computed coverage with the `luminanceToAlpha` matrix alone, which runs on straight color and drops the mask pixel's own alpha, so Firefox's `background-noodles-left-dark.svg` (its art group masked by a white→transparent gradient, the soft-vignette idiom) rendered the bottom band at full saturation. Coverage is luminance × alpha, what Chromium/Skia and Firefox compute; now 0.05–2.25 % against Chromium across the ladder (structural 0.0–1.8 %), and what remains is not the mask: the file's first rect names the mask as its `filter`, which usvg drops as an invalid reference while both browsers paint the rect unfiltered. A reduction that does exercise a luminance mask's content, `corpus/mask-luminance-radial-transform.svg` (radial gradient under a rotate-scale-translate `gradientTransform`, black→white→gray stops, nested inside a translated masked group), matches Chromium at 0.00 % / 0.000 % at 1×, 2× and 4×, with Chromium and Firefox agreeing on it.

before/after vs Chromium

**Nested masks.** A review found that a mask rect, documented as device space, was placed against the innermost layer's origin only, so a masked group inside a layer captured from a scissor origin landed displaced by that origin. The corpus hid it: the one nested mask in BuseyBench (the qwen skin-texture group) has opacity 0.14, and the >20/255-plus-erosion oracle read 0.62 % whether that group rendered or not. A high-contrast reduction of it flipped the whole face (5.58 % of pixels, 5.10 % structural), and WPT's own tests for nesting — `mask-nested-clip-path-002`, `-003`, `-panning-001`, `-panning-002` — differed from Chromium on 62–64 % of the box at 1× and 30 % at 2×. With the accumulated root-device origin, all four WPT cases match Chromium to the pixel at 1× and 2×, the reduction is at 0.13 % / 0.000 %, and the qwen group renders (removing it changes 2,903 px now, 0 before; 7,228 in Chromium). The 22-file WPT css-masking mask/nested-clip set is in the corpus with Chromium and Firefox references; the only other measured deviance there is #324's hard clip edge against an antialiased mask edge (four `content-clip` cases, 0.68 % of the box, 0.000 % structural).

nested-mask reduction and WPT mask-nested-clip-path-002: before, after, Chromium

**BuseyBench** (27 AI-generated SVG portraits, `mask` in 15 of them, `clip-path` in 24, filtered groups in all). With the layer stack plus #324/#336/#338/#339: **0.026 % mean structural difference** against Chromium 131 at 460×260, 20 of 27 files at exactly 0.000 % (22 below 0.005 %), against a 0.19 % browser-vs-browser envelope; every sheet on this page is rendered from the branch head with the review fixes. The four files above 0.05 % are blur-cap and `feBlend` residuals tracked in #325 and #332. That figure needs the whole stack; the harness now builds against exactly this PR with its clip-path and turbulence arms behind `--cfg harness_clip` / `--cfg harness_turbulence`, `validate.py` runs the documented 1×/2×/4× triple, and `compare.py --threshold 8 --exact` reports faint content the default oracle misses.

corpus at parity

### Design

- **A layer is admitted whole.** `begin_layer` reserves every image its declared effects draw through — the capture, a mask's coverage images (one, two for luminance), a filter chain's result and the scratches its passes ping-pong between, planned by the same `filter_passes()` the public chain uses — and returns `false` with everything released if any is short, so `true` promises the whole declaration instead of a filter quietly dropped past the budget. The record enumerates what it owns in one place, and a flush, `end_layer`, `reset` and a resize all keep or return that set: a masked or blurred layer open across a flush no longer loses its images (a debug assertion before; a stale-id `unwrap` on wgpu and a stale framebuffer binding on GL in release), and a reset inside a layer no longer strands its reservations until the next flush.
- **Nested masks use the root device origin.** Each record carries its store's origin in the enclosing target and the accumulated root-device origin; the mask rect is placed against the latter, so it lands where `with_mask` says at any depth.
- **A scissor clips a layer once, after its filters.** A rounded or rotated scissor set before `begin_layer` used to be re-applied inside the layer, so content was clipped before the blur and again at the composite (a half-covered edge pixel became a quarter); Chrome, Firefox and Canvas 2D under `ctx.filter` clip the result once. The store still spans the canvas for such scissors; the composite runs under the restored outer scissor.
- **`LayerEffects::default()` is `new()`** — the derived `Default` gave opacity 0, so `default().with_mask(..)` composited nothing.
- **One pass for luminance × alpha.** The mask is normalized by an ordinary draw, so the caller's image orientation never enters the parity bookkeeping. A luminance mask lands over opaque black, so the color-matrix pass's unpremultiply is the identity and the luminance it writes from premultiplied color is luminance × alpha, uncovered pixels reading coverage 0; an alpha mask lands over transparent. One `DestinationIn` draw applies the coverage, under a vertical flip when the layer was filtered.
- **The mask rect is device space** — the root's, under any nesting — because a mask is a raster captured the way the layer is; the mask image is borrowed, not owned. Luminance uses the crate's Rec. 709 weights on sRGB values (SVG's default `color-interpolation`; Skia's luma filter uses the same weights, the SVG row differs by 0.0002).
- **Two things found on the way.** A non-antialiased rect filled with an image paint drew nothing under a mirrored transform (`path_fill_is_rect` read the reversed corners as a negative extent); fixed with a GPU test. And `begin_layer`, `end_layer` and the shadow pass each wrote out the same offscreen state reset and blit; two helpers own those now, no behavior change.

### Budget

The transient-budget ladder at 1080p over the 15 BuseyBench files that reference a mask, the two reductions and the Workspace icon: 48 MiB admits every masked layer in every file and 16 of 18 render bit-identically to 256 MiB; the two others refuse 6–7 non-mask layers nested at depth 2 that the lazy reservation used to admit and then filter or not by luck (2 and 7 refusals before). At 32 MiB the pool fills its size classes at 28.7–30.4 MiB and masked layers are refused in 7 of the 15 files, so 48 MiB is the working floor for this corpus at 1080p. Table on demo-assets under `harness/pz/mask-budget-ladder.md`.

### Tests

GPU: the mask gates the layer inside its white region and hides it outside and beyond the rect; black-but-opaque hides; the two kinds read opposite things from one mask; uploaded and filtered masks keep orientation; the rect stays put under `scale(2)` with a scissor; coverage is luminance × alpha with Rec. 709 weights; the mirrored blit. From the review: masked and blurred layers survive a flush; a reset or resize inside a masked layer returns its images; a nested mask rect is root device space at depth two and three; default effects composite at full opacity; an admitted layer applies its declared filter under a three- and a four-image budget; rounded and rotated scissors clip a blurred layer one pixel inside each edge and cover an edge pixel once. Unit: images reserved during the layer and returned after (masks, chain scratches, discards), a layer short of any of them passes through whole, the scissor is absent inside a layer and present on its composite, `default()` equals `new()`, the `luminanceToAlpha` matrix, the mirrored rect. 292 GPU tests and the default suite green; fmt clean; clippy quiet in the touched code. Not addressed here: the shared untyped state stack between `save`/`restore` and layer boundaries, tracked in #335.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,14 @@ All notable changes to this project will be documented in this file.
   ramp was left transparent (or holding stale texture data) instead of the last
   stop's color, as SVG's default `spreadMethod="pad"` and Canvas gradients
   render it. Showed as a wedge cut out of the Firefox logo's flame.
+- Added layer masks: `LayerEffects::with_mask()` multiplies a layer's alpha by
+  a mask image placed in device space, using either its luminance times alpha
+  (SVG `mask`'s default `mask-type`, via the new
+  `ImageFilter::luminance_to_alpha()`) or its alpha. Masks apply after the
+  layer's filters, as in SVG, and reserve their coverage images with the
+  layer's store, so a masked layer the transient budget cannot fit passes
+  through as a whole (`begin_layer()` returns `false`) rather than composite
+  unmasked.
 - Added layers: `Canvas::begin_layer()` and `end_layer()` draw a group into an
   offscreen image and composite it back with `LayerEffects` - group opacity, so
   overlapping shapes fade as one like an SVG group, and/or an image-filter
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Run with `cargo run --example text`
 * [x] Composition modes (SourceOver, SourceIn, SourceOut, Atop, etc..)
 * [x] Image filters - gaussian blur and color matrices (the CSS filter functions), chained in a single call
 * [x] Layers - group capture with declared opacity and layer filters (Canvas 2D beginLayer/endLayer, SVG group effects)
+* [x] Layer masks - luminance and alpha (SVG mask / mask-type)
 * [x] Global alpha
 * [x] Text filling and stroking
 * [x] Text shaping
--- a/src/error.rs
+++ b/src/error.rs
@@ -42,8 +42,8 @@ pub enum ErrorKind {
     UnsupportedImageFormat,
     /// The requested operation is not supported (for example screenshot by wgpu renderer).
     UnsupportedOperation,
-    /// Acquiring another transient image (a layer backing store, filter scratch
-    /// or shadow coverage) would exceed [`crate::Canvas::set_transient_image_budget`].
+    /// Acquiring another transient image (a layer backing store, filter scratch,
+    /// shadow or mask coverage) would exceed [`crate::Canvas::set_transient_image_budget`].
     TransientImageBudgetExceeded,
 }
 
--- a/src/image.rs
+++ b/src/image.rs
@@ -550,6 +550,24 @@ impl ImageFilter {
         }
     }
 
+    /// The `feColorMatrix type="luminanceToAlpha"` conversion: the output
+    /// alpha is the input's luminance and the color channels go to zero,
+    /// which is how SVG luminance masks read their content. Uses the Rec.
+    /// 709 weights above, the ones Skia's luma filter applies for Chromium's
+    /// masks; the SVG row (0.2125, 0.7154, 0.0721) differs by at most
+    /// 0.0002. Like every color matrix it runs on straight (unpremultiplied)
+    /// color, so the luminance is the color's own, not scaled by alpha.
+    pub fn luminance_to_alpha() -> Self {
+        #[rustfmt::skip]
+        let matrix = [
+            0.0,      0.0,      0.0,      0.0, 0.0,
+            0.0,      0.0,      0.0,      0.0, 0.0,
+            0.0,      0.0,      0.0,      0.0, 0.0,
+            Self::LR, Self::LG, Self::LB, 0.0, 0.0,
+        ];
+        ImageFilter::ColorMatrix { matrix }
+    }
+
     /// CSS `opacity(amount)`; `amount` is clamped to `[0, 1]` and scales alpha.
     pub fn opacity(amount: f32) -> Self {
         let a = amount.clamp(0.0, 1.0);
@@ -788,6 +806,20 @@ mod filter_fold_tests {
     /// constant column composes in order. `brightness(0.5)` and `invert(1.0)`
     /// are both range-safe, so both directions fold, and they do not commute:
     /// brighten-then-invert is `1 - 0.5c`, invert-then-brighten is `0.5 - 0.5c`.
+    /// luminanceToAlpha writes the Rec. 709 luminance into alpha and nothing
+    /// into the color rows, so white converts to full coverage and a pure
+    /// green pixel to 0.7152 of it.
+    #[test]
+    fn luminance_to_alpha_moves_luminance_into_alpha_only() {
+        let ImageFilter::ColorMatrix { matrix } = ImageFilter::luminance_to_alpha() else {
+            panic!("luminance_to_alpha is a color matrix");
+        };
+        assert!(matrix[..15].iter().all(|c| *c == 0.0), "color rows are zero");
+        assert_eq!(&matrix[15..], &[0.2126, 0.7152, 0.0722, 0.0, 0.0]);
+        let white: f32 = matrix[15..18].iter().sum();
+        assert!((white - 1.0).abs() < 1e-4, "white is full coverage, got {white}");
+    }
+
     #[test]
     fn folding_respects_order_and_identity() {
         let bright = ImageFilter::brightness(0.5);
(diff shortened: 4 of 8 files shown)


matthargett:
Cross-checking this stack against a 26-file corpus of Firefox-ecosystem SVGs surfaced a real bug in the luminance-mask path, now fixed in 58134e2.

**Bug:** luminance coverage was computed with the `feColorMatrix luminanceToAlpha` matrix, which per feColorMatrix semantics operates on *unpremultiplied* color. That drops the mask pixel's own alpha: a white pixel fading to transparent keeps luminance 1.0, so its coverage never fades. SVG mask coverage is luminance × alpha (what Chromium/Skia and Firefox/WebRender compute — Skia's `SkLumaColorFilter` runs on premultiplied color for exactly this reason). Solid opaque-or-transparent masks — everything the suite pinned so far — can't expose the difference; any partial-alpha mask saturates to full coverage.

**Live driver:** Firefox's `background-noodles-left-dark.svg` ([SVG](https://raw.githubusercontent.com/rebeckerspecialties/femtovg/demo-assets/background-noodles-left-dark.svg)) masks its whole art group with a white→transparent gradient rect (`<g mask="url(#b)">`, mask `#b` = `<rect fill="url(#a)">`) — the common soft-vignette idiom. femtovg ignored the fade entirely and rendered the bottom band at full saturation:

before/after vs Chromium

Diff vs Chromium 149 across the zoom ladder (structural = 2-px erosion of the >20/255 diff mask, which removes edge-AA ribbons):

| zoom | before | after | before structural | after structural |
|------|--------|-------|-------------------|------------------|
| 0.7× | 5.65% | 1.29% | 4.42% | 0.42% |
| 1.3× | 18.36% | 3.35% | 16.11% | 1.76% |
| 2.1× | 29.77% | 1.71% | 27.18% | 0.07% |

The residual is edge AA plus a pre-existing top-corner clip difference unrelated to masking (present in both before and after).

**Fix:** after the DestinationIn draw of the luminance coverage, a second DestinationIn draw of the normalized mask multiplies the mask's own alpha in. It reuses the alpha-kind case's sampling convention, so it is parity-correct by construction, and adds no transient images and no shader work — one extra layer-sized, tile-local quad only in the luminance case. The regression test pins coverage at the fade midpoint (~0.5), not just the endpoints, under the mid-frame mask-capture pattern SVG integrations use.

matthargett:

|---|---|---|
| `mask` | 15/27 | #323 |

| measure | value |
|---|---|

corpus at parity

filter integration

PR #324, opened 2026-08-20
Add arbitrary-path clipping (clip-path / clip-rule) on a reserved stencil bit

matthargett:
Adds arbitrary-path clipping — `Canvas::clip_path(path, fill_rule)`, the Canvas 2D `clip()` / SVG `clip-path` + `clip-rule` semantic: intersecting under the current transform, scoped by `save()`/`restore()` (restore replays the surviving stack). `clip-path` is the most common still-missing feature in real-world SVGs, and BabylonNative's Canvas2D polyfill documents non-rectangular `clip()` as a deliberate approximation because nanovg's scissor is rectangle-only ([BabylonJS/BabylonNative#1824](https://github.com/BabylonJS/BabylonNative/pull/1824)) — this is the stencil path. Stands alone on master now that #323 has landed.

### Design: clip-tested winding on one reserved stencil bit

Both backends already carry a stencil attachment for concave fills. Bit 7 becomes the clip plane, bits 0–6 keep the winding counters, and every stencil operation stays one hardware comparison:

- Winding for a new clip accumulates only where the previous clip bit is set (`EQUAL ref 0x80 mask 0x80`, writes masked to `0x7F`), so **nesting intersects for free**; two full-target quads resolve the new bit ("clip set but no winding → clear", then "clear winding").
- Concave-fill covers switch from `NOTEQUAL 0` to `LESS 0x80`: in-clip pixels carrying winding are exactly the stencil values above `0x80`, so "visible AND covered" is one compare. Stroke counters and cleanup clears write-protect bit 7; plain draws gain a keep-only `EQUAL` test; the unclipped image-blit fast path routes through the masked path while a clip is active (the #292 precedent).
- **Clips live on the render target they were taken on.** Each entry records its target: the first clip on a target arms that target's plane, a command is gated by the clips of its own target, and popping replays only the current target's survivors. So a clip underneath a layer keeps gating the composite, a clip taken *inside* a layer (a `clip-path` on a child of a filtered or opacity group) gates the layer's content on the store's own stencil, and a popped layer-store clip goes with its store. The first version ignored a clip on a second target silently; the arm/resolve quads now span the target, which for a padded layer store is larger than the canvas.
- The plane is armed by the first `clip_path` and disarmed when the last clip restores: clip-free rendering is untouched. `set_size` re-arms an active screen clip after a resize replaces the stencil attachment; `reset()` drops its own level's clips and keeps the outer levels'; `clear_rect` clears the stencil with the colour: the whole of it when no clip is armed on the target (a plain tile clear), and only the winding bits through the stencil write mask while one is, so a clip survives the clear — a masked stencil clear is a full-target quad on Mesa's vc4 path, so the canvas says on the command which case applies — and a winding count a cover pass left behind (a fan vertex pushed past the cover quad at a cusp — the Ghostscript tiger has one at 640×480, and a colour-only clear let it blank a pixel from the second frame on) does not leak into the next frame; upstream's GL backend healed that by clearing the whole stencil each frame, wgpu never did. It stays a raw, unclipped clear (documented; Canvas 2D's clipped `clearRect` is an opaque `DestinationOut` fill, which a test pins).
- On tiled GPUs (A10X/A12 TBDR, Adreno GMEM) all of this happens in tile memory: no textures, no allocations, no render-pass breaks; a coverage-mask texture per clip level would end the pass on every clip change. Clip edges are hard (single-sample); antialiased clip boundaries are a follow-up.

### Evidence

**Clip scene** (a nonzero star clip and an even-odd donut clip over gradients and overlapping shapes, `clipdemo.svg` on demo-assets) against Chromium 131, 460×260 pivot-zoom ladder: 0.26 % / 0.54 % / 0.57 % / 0.70 % / 0.65 % of pixels at 0.5× / 1.0× / 1.1× / 1.7× / 2.5×, **0.000 % structural** at every step — the residual is the hard clip edge against the browsers' antialiased one. Before this PR the same scene was 3.4 % → 43 % across the ladder.

clip scene + Reddit, before/after vs Chromium and Firefox at 1.5x

11-step zoom ladder

**BuseyBench** (27 AI-generated SVG portraits; `clip-path` in 24 of them, `mask` in 15, filtered groups in all). Master (with #323, #336, #338, #339 and #340 landed) plus this PR: **0.026 % mean structural difference** against Chromium at 460×260, 20 of 27 files at exactly 0.000 % (22 below 0.005 %), against a 0.19 % browser-vs-browser envelope; the four above 0.05 % are the harness mapping the manual drop-shadow idiom and the SVG filter region onto whole-group effects, plus one multiply blend (#332), no clip defect in the corpus. The Reddit logo, whose `clip-path` circle was the residual after #320, is structurally exact (0.000 % after erosion at every ladder step, 0.31 % of pixels at 1×).

corpus at parity

**Equivalence.** The per-target rewrite renders bit-identically to the first version on all 32 corpus frames (no file in the corpus clips inside a layer; the new semantics are pinned by tests instead).

### Tests

wgpu: the web-platform-tests clip cases ported at the suite's pixels (`tests/wpt_clip_wgpu.rs`: `2d.path.clip.empty`, `basic.1/2`, `intersect`, `winding.evenodd.1`, `winding.1/2`, `scale`, `2d.state.saverestore.clip`/`.2`, `2d.clearRect.clip` as the `DestinationOut` form); rect clip pixel-equal to the scissor, circle corner cuts, nested intersect + restore replay, even-odd donut holes, a clipped self-intersecting concave fill proving the winding bits and clip bit never corrupt each other, clipped stencil strokes, clipped image blits routed off the fast path, an active clip surviving a resize, `reset()` dropping its level's clips and keeping outer ones, BabylonNative's nested/translated/disjoint clip scenario ([#1872](https://github.com/BabylonJS/BabylonNative/pull/1872)), a clip inside a layer, the `clear_rect` contrast, including that the clip still gates draws after the clear, and a clip popped while another target is current (`restore()` after `set_render_target`), which must reconcile its own target's plane: the survivors replayed, or the plane disarmed so a later concave fill does not read the old region as coverage. Unit: a mirrored rect still detected for the blit path; a filter pass recorded under a clip is not gated (it draws into its own target image - the OpenGL pass would otherwise be dropped). The whole open stack with this on top renders the Ghostscript tiger identically on frames one, two and three. 174 GPU tests and the default suite green; fmt/clippy clean (no new lib warnings).

### Budget

Zero new textures or heap allocations; one stencil bit; arm/disarm quads and clip resolves are a handful of stencil-only draws; nesting depth is unbounded (restore replays the stack). Winding capacity inside a clip drops from 255 to 127 self-overlaps.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1272,24 +1272,16 @@ where
         self.clear_rect(0, 0, width as u32, height as u32, Color::rgbaf(0.0, 0.0, 0.0, 0.0));
 
         // Shift device space so the captured rect's origin lands on (0, 0),
-        // exactly like the shadow pass maps its padded bbox.
+        // exactly like the shadow pass maps its padded bbox. Shadow state is
+        // a layer rendering attribute too: it applies to the layer's result
+        // at end_layer, so it must not also apply to every draw inside, or
+        // the layer's children would each cast their own shadow and then the
+        // layer would cast one more over the lot.
         let mut layer_transform = Transform2D::translation(-minx, -miny);
         layer_transform.premultiply(&state.transform);
-        self.state_mut().transform = layer_transform;
-        self.state_mut().alpha = 1.0;
-        self.state_mut().composite_operation = CompositeOperationState::default();
-        // Shadow state is a layer rendering attribute too: it applies to the
-        // layer's result at end_layer, so it must not also apply to every
-        // draw inside, or the layer's children would each cast their own
-        // shadow and then the layer would cast one more over the lot.
-        {
-            let state = self.state_mut();
-            state.shadow_color = Color::rgbaf(0.0, 0.0, 0.0, 0.0);
-            state.shadow_blur = 0.0;
-            state.shadow_offset = [0.0, 0.0];
-        }
-        if !keep_scissor {
-            self.state_mut().scissor = Scissor::default();
+        self.enter_offscreen_state(layer_transform);
+        if keep_scissor {
+            self.state_mut().scissor = state.scissor;
         }
         true
     }
@@ -1351,17 +1343,13 @@ where
         // 2D's beginLayer applies the shadow to the layer's result and SVG's
         // feDropShadow applies to a filtered group. Inside the layer the
         // shadow state was reset, so nothing has been shadowed twice.
-        let saved_transform = self.state().transform;
-        let saved_alpha = self.state().alpha;
-        self.state_mut().transform = Transform2D::identity();
-        self.state_mut().alpha = 1.0;
-
-        let mut layer_rect = Path::new();
-        layer_rect.rect(minx, miny, record.width as f32, record.height as f32);
-        self.fill_path_internal(&layer_rect, &layer_paint.flavor, false, FillRule::NonZero);
-
-        self.state_mut().transform = saved_transform;
-        self.state_mut().alpha = saved_alpha;
+        self.fill_device_rect(
+            minx,
+            miny,
+            record.width as f32,
+            record.height as f32,
+            &layer_paint.flavor,
+        );
 
         // The composite that reads the layer is recorded; its images can back
         // the next layer of this size.
@@ -1375,6 +1363,38 @@ where
         if source != capture {
             self.release_transient_image(source);
         }
+    }
+
+    /// Puts the current state into the shape every offscreen pass draws
+    /// under: `transform` as the pass's device space, full alpha, no scissor,
+    /// source-over and no shadow. The caller's `save()` holds the state this
+    /// replaces.
+    fn enter_offscreen_state(&mut self, transform: Transform2D) {
+        let state = self.state_mut();
+        state.transform = transform;
+        state.alpha = 1.0;
+        state.scissor = Scissor::default();
+        state.composite_operation = CompositeOperationState::default();
+        state.shadow_color = Color::rgbaf(0.0, 0.0, 0.0, 0.0);
+        state.shadow_blur = 0.0;
+        state.shadow_offset = [0.0, 0.0];
+    }
+
+    /// Fills a device-space rect with `paint` under the identity transform,
+    /// at full alpha and without antialiasing: the blit an offscreen pass
+    /// ends with. The scissor, composite operation and shadow state stay the
+    /// caller's - that is how a layer's composite honors the outer scissor
+    /// and casts the group's shadow.
+    fn fill_device_rect(&mut self, x: f32, y: f32, width: f32, height: f32, paint: &PaintFlavor) {
+        let saved_transform = self.state().transform;
+        let saved_alpha = self.state().alpha;
+        self.state_mut().transform = Transform2D::identity();
+        self.state_mut().alpha = 1.0;
+        let mut rect = Path::new();
+        rect.rect(x, y, width, height);
+        self.fill_path_internal(&rect, paint, false, FillRule::NonZero);
+        self.state_mut().transform = saved_transform;
+        self.state_mut().alpha = saved_alpha;
     }
 
     // Transforms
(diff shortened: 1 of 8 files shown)


matthargett:

|---|---|---|
| `mask` | 15/27 | #323 |

| measure | value |
|---|---|

corpus at parity

filter integration

src/lib.rs
```diff
@@ -1031,6 +1117,18 @@ where
...
     fn append_cmd(&mut self, cmd: Command) {
+        let mut cmd = cmd;
+        // Stencil bookkeeping commands and target switches carry no fragments
+        // to gate; clear_rect is a raw clear that neither backend clips.
+        if !matches!(
+            cmd.cmd_type,
+            CommandType::ClipFill
+                | CommandType::ClipReset { .. }
+                | CommandType::SetRenderTarget(_)
+                | CommandType::ClearRect { .. }
```
tronical, line 1242: I think `RenderFilteredImage` may have to be included here as well, otherwise filtering while a clip is active will not work.

src/lib.rs
```diff
@@ -904,14 +947,57 @@ where
...
+        *self.state_mut() = State {
+            clip_depth: inherited_depth,
+            ..State::default()
+        };
+        self.pop_clips_to(inherited_depth);
+    }
+
+    /// Drops the clips above `depth` and, when any of them gated the current
+    /// render target, re-establishes that target's plane from the survivors.
+    /// Clips of other targets need no replay: draws elsewhere never touched
```
tronical: I think this may not be correct (and thus the implementation may be incomplete). Suppose you're setting the render target to an image, then apply a clip, then switch back to the screen as render target, then restore(). I have the feeling that subsequent rendering into the image will be wrongly clipped.

src/lib.rs
```diff
@@ -1892,6 +1990,116 @@ where
...
+    /// the clips of the target underneath keep gating what is composited
+    /// back onto it. `save()` / [`restore`](Self::restore) scope clips like
+    /// the rest of the state: restoring pops the clips taken since the
+    /// matching save and replays the survivors.
+    ///
+    /// [`clear_rect`](Self::clear_rect) is a raw clear and is not clipped; a
+    /// Canvas 2D `clearRect` under a clip is a fill of the rect with an
+    /// opaque paint under [`CompositeOperation::DestinationOut`].
+    ///
+    /// The clip lives in a reserved bit of the stencil attachment both
```
tronical: Are these implementation details really relevant for users of the API?

src/lib.rs
```diff
@@ -1892,6 +1990,116 @@ where
...
+    /// backends already carry for concave fills, so it costs no textures, no
+    /// allocations, and no render-pass breaks: winding for the clip path is
+    /// accumulated only where the previous clip bit is set (nesting
+    /// intersects for free) and two bounded quads resolve the new bit.
+    /// Clip-free rendering is untouched - the plane is armed by the first
+    /// `clip_path` and disarmed when the last clip is restored. Clip edges
```
tronical: The last sentence is really a cryptic way of saying that unlike the rest of femtovg, this code way of handling paths doesn't support anti-aliasing. At the least I think this is worth simplifying and putting further up instead of the wall of text :)

src/lib.rs
matthargett: Right, and on OpenGL it was a bit worse: render_filter_pass draws through triangles(), whose clip guard tested the filter quad against the target image's freshly cleared plane, so every layer filter and blurred shadow under a clip came out blank there (wgpu's filter passes happened to ignore the flag). RenderFilteredImage is now recorded ungated, since it draws into its own target image, pinned by filter_passes_are_not_gated_by_the_active_clip in the default suite (29a7608).

src/lib.rs
matthargett: Confirmed: three new wgpu tests reproduce it (a surviving clip stayed intersected with the popped one; with no survivor the plane stayed armed and a later concave fill painted its bounding box inside the old region). A clip popped while another target is current now marks that target's plane stale, and set_render_target replays the survivors, or disarms the plane, when it becomes current again; set_size covers the screen and a released layer store drops its mark before the pool reuses it (29a7608).

src/lib.rs
matthargett: No. The doc now states the contract only; the stencil mechanics moved into a comment at the top of the function (29a7608).

src/lib.rs
matthargett: Done: the second paragraph of the doc is now 'Clip edges are not antialiased: unlike a fill or stroke edge, a pixel is either inside the clip or outside it', and the mechanics paragraph is gone from the doc (29a7608).

src/lib.rs
matthargett: This is pretty tricky, the RFC issue I filed already noted the issues in the same approximate functional area. I revised it based on all the new stuff that landed and was discussed here and in discord. https://github.com/femtovg/femtovg/issues/335

that might be the next big design thing to figure out for the next next major release

tronical: Thanks for the summary ticket. I tried to err shorten it further ;). I think this is the last thing to decide on for this PR.

src/lib.rs
tronical: Thanks :)

src/lib.rs
matthargett: Kept this in #324. Popped clips now mark their own target dirty and replay lazily before the next draw on that target; consecutive restores coalesce. Target switches, resize/reallocation, filtering, suppression, and layer-store reuse have focused coverage. #335 remains the separate layer-nesting design issue.

Issue #310, opened 2026-08-04
Glyph artifacts on VideoCore IV (GLES2): atlas neighbor bleed at margin 1, and frozen stray pixels from the >92 px direct path


fuzzy1450:
Two related glyph-rendering defects observed on VideoCore IV hardware, isolated with per-build before/after measurements on real framebuffer captures. Desktop GL renders the identical pipeline clean; both defects are specific to the embedded GPU path, and both have one-line mitigations we now run in production. Filing as one issue because they present as a single user-visible symptom; happy to split if you prefer.

## Environment

- Raspberry Pi 3 Model B Rev 1.2 (VideoCore IV; the `vc4` Mesa driver, GL_RENDERER V3D 2.1 class), Mesa **26.0.1**, kernel 6.12.61, GLES2 via DRM/KMS + GBM + EGL
- femtovg **0.25.1**, `default-features = false, features = ["swash"]` — driven by Slint 1.17.1's femtovg renderer (LinuxKMS backend), 1920×1080 @ dpi 1.0
- Font: DejaVu Sans Bold; the affected headline renders at **96 px** via `fill_glyph_run` (positioned glyphs)
- All captures are `glReadPixels` readbacks of the presented frame (we also confirmed the artifacts independently in the scanout buffer via `/dev/mem`) — real pixels, not camera or panel effects
- Desktop control: same harness on Windows desktop GL (NVIDIA RTX 4080) — clean

## Symptom

Tiny pure-white specks (1–2 px, exactly the glyph fill color) sit 2–12 px off the strokes of large text, ~3 per big-type scene, at coordinates that are **frozen frame-to-frame**. Separately, frames also showed transient multi-speck clusters (up to 14 in a frame). Context shot (96 px headline, speck barely visible at this scale — the crops below are ×8):

![context](https://gist.githubusercontent.com/fuzzy1450/fe65f129a53dd0a7c618b599b3121e48/raw/0cfe86cd1299ab68e3d27421321b167ca281a87b/femtovg-vc4-context-wide.png)

## Measurements

Detector: connected-component analysis over framedumps — a "speck" is a component of ≤6 px in glyph-fill color whose distance to the nearest large component is 2–12 px (script below). Three builds, 4-minute capture bursts of a cycling scene set, one variable changed at a time:

| build | change | specks / frames | affected frames |
|---|---|---|---|
| stock 0.25.1 | — | 226 / 364 | 54 (14.8 %) |
[...]

PR #330, opened 2026-08-28
Zero-fill the glyph margin ring in swash atlas uploads

matthargett:
The atlas texture starts uninitialized, and the swash route uploads only the sampled rect — glyph + `GLYPH_PADDING` — at an offset of `GLYPH_MARGIN` into the cell. The margin ring is never written. So the first texel outside every swash glyph's sampled area is undefined VRAM, and on a GPU whose interpolated texture coordinates drift at quad edges, NEAREST selects it and paints whatever it holds. That is #22's defect class (uninitialized atlas memory around glyphs) reintroduced for the swash path; the non-swash path route clears the whole cell and is unaffected.

This is the plainest candidate mechanism for #310's finding 1 — fill-colored specks around swash text on VideoCore IV, a GPU where Mesa [#4359](https://gitlab.freedesktop.org/mesa/mesa/-/issues/4359) independently documents one-texel texture over-reads. Desktop GPUs don't drift, so the undefined ring never shows there.

Fix: build the full allocated cell — glyph at the padding+margin offset, every other texel zeroed — and upload it at the cell origin. No constant changes, no memory growth, no layout change; the sampled rect and quads are untouched, so output is bit-identical on hardware with exact interpolation. Also updates the `GLYPH_MARGIN` comment, which still described bilinear bleed although the atlas has been NEAREST-sampled since 62e5be6 (#310 flagged the stale comment as well).

Tests pin the cell contract: full-allocation coverage, blank padding and margin rings, ink at the padding+margin offset, and the color-glyph variant.

The #310 reporter has offered to A/B candidate fixes on the affected hardware; I've asked in the issue for a run of this patch at stock constants.

### DIFF
--- a/src/text.rs
+++ b/src/text.rs
@@ -26,10 +26,13 @@ pub use textlayout::*;
 // This padding is an empty border around the glyph’s pixels but inside the
 // sampled area (texture coordinates) for the quad in render_atlas().
 const GLYPH_PADDING: u32 = 1;
-// We add an additional margin of 1 pixel outside of the sampled area,
-// to deal with the linear interpolation of texels at the edge of that area
-// which mixes in the texels just outside of the edge.
-// This manifests as noise around the glyph, outside of the padding.
+// We add an additional margin of 1 texel outside of the sampled area.
+// The atlas is NEAREST-sampled (since 62e5be6), so no filter tap blends
+// across the boundary; the margin is a guard band for GPUs whose
+// interpolated texture coordinates drift at quad edges and select a
+// neighbouring texel instead (issues #22, #310). Margin texels are
+// written blank at upload time - the atlas texture itself starts
+// uninitialized, so an unwritten margin would hold undefined data.
 const GLYPH_MARGIN: u32 = 1;
 
 const TEXTURE_SIZE: usize = 512;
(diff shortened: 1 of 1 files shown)


tronical (commented):
Hmm wouldn't it be easiest to just zero initialize the entire atlas texture on creation? And don't we do that already?

matthargett:
> Hmm wouldn't it be easiest to just zero initialize the entire atlas texture on creation? And don't we do that already?

not at my computer (using github app on my phone), but IIRC on the GL backend the texture is allocated with tex_image_2d(..., None), contents undefined. there was a comment above the upload source line tht said similar, and the debug-inspector red fill exists to make that visible (but only in debug builds). wgpu _does_ zero-initialize per spec, and the desktop GL drivers I tested on Windows 10/11 and Linux 6.x all hand out zeroed pages in practice, which is probably only why only this technically-compliant-but-still-deviant embedded GL case shows it. I'm betting if we have the user run a debug build, the red pixels will show up more readily.

Zeroing the whole texture at creation should work,  but the simple version has some complications: clear_rect is deferred to flush while swash uploads are immediate, so a same-frame bake gets wiped. the debug-inspector code comments on this situation and uses an immediate upload instead. that seems to leave us with a one-time ~1MB zeroed upload per texture. we can minimize this by making the atlas pool by gray8 instead of rbga8, but the user isn't complaining about RAM use or memory bandwidth overhead yet. I have my own deployment targets on embedded boards for my WAMR + femtovg + wgpu runtime in the back of my mind also.

I can commit the clear_rect in this PR as well, but I'd advocate for keepin the per-cell coverage either way: it's the invariant that stays true if cells ever get recycled within a texture (eg if the atlas grows eviction someday like in some animated SVG approaches), it matches what the path route already does for its cells, and it writes fewer bytes. 

if I'm misunderstanding, or I'm gold-plating for futures/footguns that don't match with the project, tell me to just do creation-init only. that seems fine for the current codebase, so it's dealer's/maintainer's choice from my perspective :) I'd still like to hear from the user on the A/B to see how simple we can make the fix.

tronical:
Great, if wgpu guarantees zero initialization then I think it makes even more sense to just temporarily allocate that heap vec and use glTexImage2D() to clear the texture and then we're done :)

matthargett:
Superseded by #331, which takes the creation-time zero-init approach from the discussion above. Closing.

PR #343, opened 2026-09-20
Keep wgpu pipelines across flushes within a frame

zoujiaqing:
A canvas flushes several times per frame (glyph atlas uploads, filtered layers, the screen pass). render() retained only the pipelines accessed since the previous flush, so pipelines used by a later pass were evicted and recreated on the next frame, most visibly while scrolling text. Sweep unaccessed pipelines only once the cache holds more than 64 states.


tronical (approved):
Good idea, looks like an improvement:)

matthargett:
good catch! I might be reading it incorrectly, but it seems like there is a boundary condition around the limit we can retain unused pipelines for the lifetime of the program.

Can you add some unit/integration tests that:
* count pipeline creations and resident entries at 63/64/65 states
* test alternating glyph, clipped-layer, filter, and screen flushes (after merging from current master) 
* bonus points: Measure frame-time tails and driver memory on an actual WGPU device.

I'm betting that alternating large and small flushes can repeatedly recreate dozens of pipelines. While that problem is largely pre-existing, this PR exacerbates it because its extending residency. lmk if you disagree or need any help! :D

PR #345, opened 2026-09-20
Add an affine transform to image paints

zoujiaqing:
Paint::set_image_transform / with_image_transform map an image paint's own coordinate space into user space before the canvas transform, the image counterpart of set_gradient_transform. This is what SVG patternTransform and Canvas 2D CanvasPattern.setTransform need: with REPEAT_X | REPEAT_Y images it draws sheared or non-uniformly scaled repeating patterns that Paint::image's centre/size/angle cannot express.

The axis-aligned blit shortcut already derives its texture coordinates from the paint matrix, so it honours the transform unchanged.


matthargett:
A good change, but it exposes a latent problem more readily then before: a custom path that becomes an axis-aligned rectangle under a rotated or sheared canvas transform takes the shortcut, then disagrees with the normal fill. Your new image-paint transform can shear a pattern over an ordinary rectangle, while the shortcut’s eligibility check around line 395 in paint.rs still checks only angle == 0. 

on my m4 macbook, a sheared repeating pattern rendered red through the non-AA rectangle shortcut and green through the normal fill. adding a clip changed that same interior pixel from red to green. The shortcut at line 3414 in lib.rs computes texture coordinates at two corners and constructs the other two, which is invalid for shear. oopz.

I think this feature should account for it before merge—either calculate all four UV corners or exclude non-axis-aligned image transforms from the shortcut, with tests for both the old-API repro and ordinary sheared patterns.

lmk if you'd like to handle it in this PR, or I can do it in a separate PR before this one merges :D

